PR #1, opened 2022-09-15
Initial package setup

rth:
Mostly boilerplate code to set up the package adapted from pytest-pyodide, matplotlib-pyodide etc.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,4 @@
+[flake8]
+extend-ignore = E203, E402, E501, E731, E741, B950, W503
+select = C,E,F,W,B,B9
+extend-immutable-calls = typer.Argument, typer.Option
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,32 @@
+name: main
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: main-${{ github.head_ref || github.run_id }}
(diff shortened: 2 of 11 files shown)


rth:
It does look like we have some misconfiguration for pre-commit, because it does find things to fix in code I copy-paster from the Pyodide repo. Where pre-commit should have been running on them.

hoodmane:
I think @henryiii specifically excluded the externals directory based on the idea that we didn't want to change code copied out of other projects.

rth:
Yeah but for instance, in https://github.com/pyodide/micropip/pull/1/commits/851b14b7a73f9e0a814a10c8c72b9b795518498d triggered by pre-comit there is a linter change in `micropip/_micropip.py` and in `tests/test_micropip.py`. So it's not just externals. With a monorepo it's  hard to make sure that path regexes are correct everywhere, because there are so many things...

hoodmane:
Probably would be good to revert the changes to the externals directory.

rth:
I did in earlier commits..

hoodmane:
> I did in earlier commits..

Sorry!

Recently I have also been getting the following error:
```
/home/hood/.cache/pre-commit/repow1xom7p_/py_env-3.10/lib/python3.10/site-packages/numpy/__init__.pyi:636: error: Positional-only parameters are only supported in Python 3.8 and greater  [syntax]
```

PR #3, opened 2022-09-17
Enable test CI + more packaging improvement

rth:
Enable test CI (mosty adapted from pytest-pyodide) for now only running tests in Chrome.

Also other packaging improvements (e.g. put the contents of `setup.py` into `pyproject.toml`).


rth:
OK, actually running tests on the host depends on `src/py/` in Pyodide. For now, I'm going to vendor a minimal subset needed to pass the test suite but we should publish `src/py` to PyPI avoid this https://github.com/pyodide/pyodide/issues/1839

PR #4, opened 2022-09-19
Remove vendored source code

ryanking13:
This PR adds more compatibility codes to drop vendored Pyodide source code.

- to_js
- get_dynlibs
- wheel_dist_info_dir

`wheel_dist_info_dir` is copied from the original Pyodide source code, while other two are dummy functions.


conftest.py
```diff
@@ -0,0 +1,3 @@
+import sys
+
+sys.path.append("micropip")
```
rth, line 3: Are you sure about this? Normally it shouldn't be necessary.

rth (approved):
Thanks @ryanking13 this is very clearly a nice improvement!

Though I think it would still make sense to release https://github.com/pyodide/pyodide/issues/1839 and depend on `pyodide-py` for tests, so we are sure we are using exactly the same version of canonicalize_name etc

conftest.py
ryanking13: I'm not sure why but that was needed for me to run test locally. I'll remove that file and let's see if CI complains.

ryanking13:
> Though I think it would still make sense to release [pyodide/pyodide#1839](https://github.com/pyodide/pyodide/issues/1839) and depend on `pyodide-py` for tests, so we are sure we are using exactly the same version of canonicalize_name etc

Yes that would be optimal.

conftest.py
ryanking13: Okay, it works. Maybe my local setting was wrong.
(The test failure is about codecov upload)

hoodmane: > (The test failure is about codecov upload)

This happens a lot, it's a bit annoying. If it's going to happen regularly, maybe we can figure out a way to not fail the test when the upload fails.

ryanking13:
@rth It seems like I don't have permission to merge this (also in `matplotlib-pyodide`).

rth:
> It seems like I don't have permission to merge this (also in matplotlib-pyodide).

Should be fixed now. CI is failing here though.

ryanking13:
> > It seems like I don't have permission to merge this (also in matplotlib-pyodide).
> 
> Should be fixed now. CI is failing here though.

The CI failure happens in codecov upload, so I think it is okay to merge this. I'm not sure why codecov shows a lot of 404 errors.

rth:
Thanks!

There is a setting to not fail on upload errors as far as I remember.

Issue #11, opened 2022-07-15
TST micropip test_install_keep_going fails in some environments


rth:
Running,
```
$ pytest packages/micropip/ -k test_install_keep_going 
```
fails with,
```
packages/micropip/test_micropip.py:536: in test_install_keep_going
    with pytest.raises(ValueError, match=msg):
E   Failed: DID NOT RAISE <class 'ValueError'>
```
in some environments (e.g. in CI https://github.com/pyodide/pyodide/pull/2871 or for me locally) but passes on main in the docker image. So it sounds like it's not fully deterministic.

PR #13, opened 2022-09-29
Fix keep going test

ryanking13:
Close: #11

The test was failing because when running host test, micropip accepts manylinux wheels.

### DIFF
--- a/tests/test_micropip.py
+++ b/tests/test_micropip.py
@@ -89,8 +89,10 @@ def make_wheel_filename(name: str, version: str, platform: str = "generic") -> s
         platform_str = "py3-none-any"
     elif platform == "emscripten":
         platform_str = f"{cpver}-{cpver}-{_platform()}"
-    elif platform == "native":
+    elif platform == "linux":
         platform_str = f"{cpver}-{cpver}-manylinux_2_31_x86_64"
+    elif platform == "windows":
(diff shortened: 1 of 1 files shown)


tests/test_micropip.py
```diff
@@ -541,15 +543,14 @@ def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
...
 @pytest.mark.asyncio
 async def test_install_keep_going(mock_fetch: mock_fetch_cls) -> None:
     dummy = "dummy"
     dep1 = "dep1"
     dep2 = "dep2"
     mock_fetch.add_pkg_version(dummy, requirements=[dep1, dep2])
-    mock_fetch.add_pkg_version(dep1, platform="native")
-    mock_fetch.add_pkg_version(dep2, platform="native")
+    mock_fetch.add_pkg_version(dep1, platform="windows")
+    mock_fetch.add_pkg_version(dep2, platform="windows")
```
rth, line 553: Why windows though? Is the assumption that we won't run this on windows?

rth (commented):
Thanks!

tests/test_micropip.py
ryanking13: > Is the assumption that we won't run this on windows?

Right, but probably I should just use a invalid platform string.

PR #20, opened 2022-10-26
Run tests with local micropip

ryanking13:
This makes tests run using local micropip, instead of using one from downloaded Pyodide release.

### DIFF
--- a/tests/test_micropip.py
+++ b/tests/test_micropip.py
@@ -195,17 +195,48 @@ def mock_fetch(monkeypatch, mock_importlib, wheel_base):
     return result
 
 
+@pytest.fixture(scope="function")
+def install_wheel(selenium):
+    wheel_dir = Path(__file__).parent / "wheel"
+    wheel_files = list(wheel_dir.glob("*.whl"))
+
+    if not wheel_files:
+        pytest.exit("No wheel files found in wheel/ directory")
+
+    wheel_file = wheel_files[0]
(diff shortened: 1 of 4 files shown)


tests/test_micropip.py
```diff
@@ -200,12 +200,25 @@ def selenium_standalone_micropip(selenium_standalone):
...
-        await pyodide.loadPackage("micropip");
-        pyodide.runPython("import micropip");
-        """
-    )
+
+    wheel_dir = Path(__file__).parent.parent / "wheel"
+    wheel_files = list(wheel_dir.glob("*.whl"))
+
+    if not wheel_files:
+        pytest.exit("No wheel files found in wheel/ directory")
```
rth, line 208: Good point, thanks!

Should we just create a wheel at runtime here (maybe using their [programmatic API](https://pypa-build.readthedocs.io/en/stable/api.html#build.ProjectBuilder)? Otherwise, it would be annoying to use these tests locally if one has to always re-generate wheels before running tests.

ryanking13: > Should we just create a wheel at runtime here (maybe using their [programmatic API](https://pypa-build.readthedocs.io/en/stable/api.html#build.ProjectBuilder)? Otherwise, it would be annoying to use these tests locally if one has to always re-generate wheels before running tests.

Thanks for the info! I'll try to use that API.

ryanking13: Okay, I managed to make it work.

Issue #16, opened 2022-10-15
Enhancements ideas from JupyterLite


rth:
As summarized in https://github.com/pyodide/pyodide/issues/3093#issuecomment-1245786852 by @bollwyvl there are a number of features in piplite/JupyterLite built on top of micropip, some of which could make sense to move upstream, now that this package is standalone, 

> What piplite/jupyterlite hacks are wanted? Happy to shed some stuff!
>
> Even if all our hacks weren't desirable, it would be lovely to have a few more documented ways to customize micropip, rather than trying to solve some things Once And For All... presenting the two together:
>
>   1. handling multiple package sources
>        1. we [check](https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/py/piplite/piplite/piplite.py?rgh-link-date=2022-09-13T18%3A16%3A36Z#L65) for a custom Warehouse (or [warehouse-like](https://github.com/pypi/warehouse/blob/main/warehouse/legacy/api/json.py?rgh-link-date=2022-09-13T18%3A16%3A36Z#L96) [file format](https://jupyterlite.readthedocs.io/en/latest/_static/pypi/all.json)) for custom/offline packages
>        2. notionally, micropip.add_pypi_json_resolver(key: str, get_pypi_json: Callable[[str, dict], PackageReleases])
>  1.  disabling package sources
>        1. for offline (but still dynamic) installations, it's nice to have some guardrails in place to avoid calling out to Big PyPI
>        2. notionally, micropip.remove_pypi_json_resolver(key: str)
>  1. specifying no-op packages to be considered installed
>        1. not handled in piplite, but rather in separate build packages
>        2. notionally, micropip.add_noop_package(package_name: str)
>
> Some things that we hack outside of piplite, so maybe they belong in core:
> 
>  4.  specifying patches that need to occur at import time
>        1. we [un-lazily load and patch](https://github.com/jupyterlite/jupyterlite/blob/main/packages/pyolite-kernel/py/pyolite/pyolite/patches.py?rgh-link-date=2022-09-13T18%3A16%3A36Z) some heavy hitters (matplotlib, oy!)
>        2. this is really more of a loader-time thing, so maybe should be a pyodide API
>        3. we tried to use some off-the-shelf stuff, but at some point it stopped working
>  5.  specifying mock packages
[...]

PR #22, opened 2022-11-02
Option to skip selected dependencies

joemarshall:
Many packages include dependencies which are only used in some cases. e.g. I've got some packages which include a load of jupyter server stuff as deps, which can be ignored in pyodide. 

This patch lets you skip some dependencies if you don't want to get them, whilst still getting other deps with micropip.

### DIFF
--- a/micropip/_micropip.py
+++ b/micropip/_micropip.py
@@ -288,6 +288,7 @@ class Transaction:
     deps: bool
     pre: bool
     fetch_kwargs: dict[str, str]
+    skip_deps: list[str]
 
     locked: dict[str, PackageMetadata] = field(default_factory=dict)
     wheels: list[WheelInfo] = field(default_factory=list)
@@ -346,6 +347,9 @@ class Transaction:
         See PEP 508 for a description of the requirements.
         https://www.python.org/dev/peps/pep-0508
         """
+        if req.name in self.skip_deps:
+            # don't install any dependencies we've been told to ignore
+            return
         for e in req.extras:
             self.ctx_extras.append({"extra": e})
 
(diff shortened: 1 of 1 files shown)


ryanking13:
Thanks, @joemarshall! I am a bit undecided about this PR. We have `deps` option which exists for a similar purpose. But I agree that it would be convenient to have options to "ignore" specific packages if a package has tons of dependencies. So if @rth or @hoodmane is okay with this. I have no objections to merging this.

rth:
Yes, this would be useful, thanks @joemarshall ! This was also requested by @bollwyvl in point 3 in https://github.com/pyodide/micropip/issues/16. The only thing I'm not certain is in terms of API whether it would be better to have,
 1. this `skip_deps` passed to each `install` call
 2. a API such as `micropip.add_noop_package(package_name: str | list[str])` as proposed in the linked issue. Which could be provided once. Later, it could also take as an optional argument whether the package needs to be mocked (in case it's a nonrelevant package but imported somewhere in the dependencies).  The name of this function remains to be determined.
 
 WDYT?

ryanking13:
`add_noop_package` API looks fine to me, as third-party Pyodide-based application developers can specify packages that can be ignored, and users who use it doesn't have to be aware of what is happening in the background (I think it is why jupyterlite team wants this feature?).

joemarshall:
Ooh, that is a super idea. Maybe even add_mock_package, with an option to pass in a module object

joemarshall:
I have a neat little class which makes a module from itself, with class members becoming module members, with add_mock_package, I could just chuck those straight into micropip and they'd only get into the sys namespace when micropip install was called on something with them as a dependency

PR #26, opened 2022-11-17
Add add_mock_package

joemarshall:
Added a method to micropip to allow easy mocking of packages (for use in jupyterlite / pyodide)

It takes package name and version, along with a list of implemented modules and optionally their code if needed.


micropip/_micropip.py
```diff
@@ -692,3 +695,72 @@ def _list():
...
+        for n, s in modules.items():
+            mf.write(f"Provides:{n}\n")
+
+            if s is None:
+                s = ""
+            s = dedent(s)
+            path_parts = n.split(".")
+            dir_path = Path(site_packages, *path_parts)
+            os.makedirs(dir_path, exist_ok=True)
+            init_file = dir_path / "__init__.py"
```
rth, line 764: Do we actually need to create these physical files?

I was thinking we could maybe directly initialize `sys.modules` with the mapping from `modules: dict[str, ModuleType | str]`? That way it's also easy to use `MagicMock` as a module or any other custom object defined at runtime to make a given dependency work.

 Though yes, the API to create modules dynamically via [`importlib.util.module_from_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.module_from_spec)  or by [inheriting from ModuleType](https://stackoverflow.com/a/67813054/1791279) isn't very straightforward.

hoodmane: Maybe making another custom Finder/Loader would be a good way to go. The API for that isn't too bad.

joemarshall: I actually wrote it initially using mock objects and putting them into sys.modules, but I reverted to writing them out to files. I did it like that because it is really simple and reliable, things like file paths etc. just work. It also works nicely in terms of loading modules, in that init code is only run on actual import. Oh and in the case that you're running somewhere with persistent file system, your mock modules continue to exist on reload of python.

It's a pretty minor change - you just create a module and exec the code into it in the loader, I can put that back in if you reckon the possibility of using mock objects makes sense.

joemarshall: Actually I think rather than take an object it needs to take a callable which takes a module object and does whatever needs to be done on it to make it. I.e. the equivalent of loader.exec_module

That way you can mock module initialisation correctly.

joemarshall: One further thought on this - at some point I want to implement some kind of mega wheel option, to bundle a set of wheels and all deps into a single wheel with no deps. For that use case keeping persistent mock modules would be nice. If it's okay I'd like to keep that as an option in micropip?

It would also be nice for any point that pyodide is run in a persistent file system. 

I'm thinking that a flag "persistent" could be added (which only works for string modules) and then the behaviour with finder and loader could be used in the case that the flag isn't set.

ryanking13:
> It would also be nice for any point that pyodide is run in a persistent file system.

I think this is reasonable, as having a physical file would be a most easy way to keep the state of installed packages or to bundle packages (for example, in `pyodide-pack`) for later use.

We can change the implementation of `add_mock_package` later if we find a better way to deal with this, but I'm +1 with this approach for now.

micropip/_micropip.py
```diff
@@ -692,3 +695,72 @@ def _list():
...
+
+       ``
+
+    Parameters
+    ----------
+    name : ``str ``
+
+        Package name to add
+
+    modules : ``Optional(dict) ``
```
ryanking13, line 733: docstring for `version` parameter is omitted

micropip/_micropip.py
```diff
@@ -692,3 +695,72 @@ def _list():
...
+        Dictionary of module_name:string pairs.
+        The string contains the source of the mock module or is blank for
+        an empty module.
+    """
+    if modules is None:
+        # make a single mock module with this name
+        modules = {name: ""}
+    # make empty mock modules with the requested names in user site packages
+    site_packages = Path(site.getusersitepackages())
+    metadata_file = site_packages / (name + "-" + version + ".egg-info")
```
ryanking13, line 744: Perhaps use `dist-info` directory instead?

micropip/_micropip.py
```diff
@@ -692,3 +695,72 @@ def _list():
...
+        )
+        for n, s in modules.items():
+            mf.write(f"Provides:{n}\n")
+
+            if s is None:
+                s = ""
+            s = dedent(s)
+            path_parts = n.split(".")
+            dir_path = Path(site_packages, *path_parts)
+            os.makedirs(dir_path, exist_ok=True)
```
ryanking13, line 763: Maybe we should raise error if a directory exists, because it would mean that the package is already installed somehow.

micropip/_micropip.py
```diff
@@ -692,3 +695,72 @@ def _list():
             source_ = pkg_source
         packages[name] = PackageMetadata(name=name, version=version, source=source_)
     return packages
+
+
+def add_mock_package(name, version, *, modules=None):
```
ryanking13, line 700: Could you please add type hints?

micropip/_micropip.py
```diff
@@ -692,3 +695,72 @@ def _list():
...
+
+            if s is None:
+                s = ""
+            s = dedent(s)
+            path_parts = n.split(".")
+            dir_path = Path(site_packages, *path_parts)
+            os.makedirs(dir_path, exist_ok=True)
+            init_file = dir_path / "__init__.py"
+            with open(init_file, "w") as f:
+                f.write(s)
```
ryanking13, line 766: I think we should keep the list of mocked modules.

micropip/_micropip.py
joemarshall: done

micropip/_micropip.py
joemarshall: Added list_mock_modules instead - which uses module installer metadata to list modules it has installed

joemarshall:
@ryanking13 @rth I've added support for non-persistent mocks, and made the tests work in pyodide. I think I've fixed the things from the reviews above.

There's a bit more code in there to do the in-memory mocking, which is probably worth a look over before merging.

joemarshall:
Oh and it supports listing and removing mocks also btw.

micropip/_micropip.py
```diff
@@ -692,3 +698,166 @@ def _list():
...
+            s = dedent(s)
+            path_parts = n.split(".")
+            dir_path = Path(site_packages, *path_parts)
+            dir_path.mkdir(exist_ok=True, parents=True)
+            init_file = dir_path / "__init__.py"
+            file_list.append(init_file)
+            with open(init_file, "w") as f:
+                f.write(s)
+        with open(installer_file, "w") as f:
+            f.write("micropip mock package")
```
ryanking13, line 813: Maybe move these hard coded INSTALLER types into a variables?

micropip/_mock_package.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def find_spec(self, fullname, path=None, target=None):
+        if fullname not in _mock_modules.keys():
+            return None
+        spec = importlib.util.spec_from_loader(fullname, self)
+        return spec
+
+
+_finder = _MockModuleFinder()
+sys.meta_path = [_finder] + sys.meta_path
```
ryanking13, line 79: I'm not sure we want an importing micropip to have a side effect of modifying `sys.meta_path`, though it does not have any effect when no mock packages are registered.

Perhaps modify sys.meta_path when `add_in_memory_distribution` is first called?

tests/test_micropip.py
```diff
@@ -934,3 +934,181 @@ def test_check_compatible(mock_platform, interp, abi, arch, ctx):
...
+    assert importlib_version("test_1") == "1.0.0"
+
+
+def test_persistent_mock(monkeypatch, capsys):
+    import site
+    from importlib.metadata import version as importlib_version
+
+    from micropip._micropip import add_mock_package
+
+    with TemporaryDirectory() as tmpdirname:
```
ryanking13, line 986: There is a [`tmp_path`](https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html) fixture for this purpose.

tests/test_micropip.py
```diff
@@ -934,3 +934,181 @@ def test_check_compatible(mock_platform, interp, abi, arch, ctx):
...
+        if d.name == "micropip_test_bob":
+            found_bob = True
+    assert found_bob is True
+    assert (
+        importlib.metadata.distribution("micropip_test_bob").name == "micropip_test_bob"
+    )
+    # check package removes okay
+    _micropip.remove_mock_package("micropip_test_bob")
+    del micropip_bob_mod
+    try:
```
ryanking13, line 1106: Maybe use `with pytest.raises(...)` instead?

tests/test_micropip.py
```diff
@@ -934,3 +934,181 @@ def test_check_compatible(mock_platform, interp, abi, arch, ctx):
...
```
ryanking13, line 1058: Same here (`pytest.raises`)

ryanking13 (approved):
Thanks, @joemarshall. Having a persistent parameter seems good to me. Overall LGTM, if @rth or @hoodmane is also +1 with this, I'll merge.

Could you please also update the [changelog](https://github.com/pyodide/micropip/blob/main/CHANGELOG.md)? Probably we can follow the format of [KeepAChangelog](https://keepachangelog.com/en/1.0.0/)

joemarshall:
Okay, done those bits and pieces now and changelog. Good spot about the sys.meta_path thing, that was silly.

micropip/_micropip.py
```diff
@@ -692,3 +698,170 @@ def _list():
...
+
+       ...
+       ...
+
+       {
+          "mylovely_module": init_fn
+       }
+       ``
+
+    Parameters
```
rth, line 751: I think there is an indentation issue in this docstring, but we can fix it once we start actually building docs for it.

rth (approved):
Thanks a lot for this work @joemarshall  and @ryanking13 for the review!

ryanking13:
I guess we should release 0.1.1 (or 0.2.0) with this. Though probably we need to start building docs for micropip before making a release.

PR #28, opened 2022-11-28
CI Use pyodide/pyodide-actions

ryanking13:
Use `pyodide/pyodide-actions` to download Pyodide and install browser runtimes.


rth (approved):
Looks much better indeed! Thanks!

PR #33, opened 2022-12-09
Handle urls with search params

hoodmane:
- [x] Add a test
- [x] Changelog


rth (commented):
Merged with upstream/main to fix CI.

Indeed changelog and maybe a test needed, otherwise LGTM. Thanks!
Worth mentioning that as far as I remember, we also tested that urlparse works with arbitrary input generated by hypothesis.

rth:
OK I added a changelog, a test, and checked that the test failed before this fix. Merging so we can make the 0.2 release.

PR #34, opened 2022-12-10
Fixed wheel choice by interpreter version tag

### LINKED ISSUE #32: `micropip.install` is mishandling wheel tags
## 🐛 Bug

`micropip.install` does not reliably select the best-matching tagged wheel.

### To Reproduce

1. Open [the official website's Pyodide REPL](https://pyodide.org/en/stable/console.html)

2. Use `micropip` to install a package with multiple pure Python 3 wheels. For example, [my `compose` package](https://pypi.org/project/compose/) has three (one for older Pythons, one for 3.5+ with `async`/`await` features added, and one for 3.8+ to use native PEP-570 positional-only arguments):

    ```python
    >>> import micropip
    >>> await micropip.install('compose')
    ```

3. Try importing `acompose`. If the `py2.py30`-tagged wheel got installed, this will fail:

    ```python
    >>> from compose import acompose
    ImportError: cannot import name 'acompose' from 'compose' (/lib/python3.10/site-packages/compose.py)
    ```

### Expected behavior

The wheel tagged `py38` gets installed (since Pyodide in that REPL is 3.10 at this time), import of `acompose` succeeds, and inspecting the signature of `acompose.__call__` shows that PEP-570 positional-only syntax is being used:

```python
>>> from compose import acompose
>>> help(acompose.__call__)
Help on function __call__ in module compose:

async __call__(self, /, *args, **kwargs)
    Call the composed function.

>>> 
```

mentalisttraceur:
Fixes #32 .

There seem to be *unrelated* test failures *already* in the automated test suite, but those failures are the same both before and after my patch, and the tests I added (and the one test I changed to factor out some now-reused code) are all passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for adding mock packages, for use where something is a dependency and you don't need it, or you need only a limited subset of the package. This is done using `micropip.add_mock_package`, `micropip.remove_mock_package` and `micropip.list_mock_packages`. Packages installed like this will be skipped by dependency resolution when you later install real packages.
 
+### Fixed
+
+- When multiple compatible builds for a package exist, the best
+  build is now installed, as determined by the order of tags in
+  [`packaging.tags.sys_tags`](https://packaging.pypa.io/en/latest/tags.html#packaging.tags.sys_tags).
+  For example, if a package has two pure Python wheels, one tagged `py30` and
+  another tagged `py35`, the `py35` wheel will now always get installed.
+  [#34](https://github.com/pyodide/micropip/pull/34)
+
 ## [0.1.0] - 2022/09/18
 
 Initial standalone release. For earlier release notes, see
--- a/micropip/_micropip.py
+++ b/micropip/_micropip.py
@@ -89,13 +89,27 @@ class WheelInfo:
             parsed_url=parsed_url,
         )
 
+    def best_compatible_tag_index(self) -> int | None:
+        """Get the index of the first tag in ``packaging.tags.sys_tags()`` that this wheel has.
+
+        Since ``packaging.tags.sys_tags()`` is sorted from most specific ("best") to most
+        general ("worst") compatibility, this index douples as a priority rank: given two
+        compatible wheels, the one whose best index is closer to zero should be installed.
+
+        Returns
+        -------
+        ``int | None``
+            The index, or ``None`` if this wheel has no compatible tags.
+        """
+        for index, tag in enumerate(sys_tags()):
+            if tag in self.tags:
+                return index
+        return None
+
     def is_compatible(self):
         if self.filename.endswith("py3-none-any.whl"):
             return True
-        for tag in sys_tags():
-            if tag in self.tags:
-                return True
-        return False
+        return self.best_compatible_tag_index() is not None
 
     def check_compatible(self) -> None:
         if self.is_compatible():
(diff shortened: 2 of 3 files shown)


mentalisttraceur:
Local testing was a struggle. I couldn't find any documentation on it, but between the various errors and tracebacks and a peek at the definition of the GitHub PR testing action, I figured out that I needed to

1. create a `dist` directory in the repo root,
2. download and unpack a Pyodide release into it (so it should look like `./dist/pyodide/...`), then
3.  create a `conftest.py` with a `DIST_PATH` which is a `pathlib` object that points to that `dist` folder, and finally
4. run `pytest` with `--dist-dir=./dist/`.

That got *most* of the tests running locally.

Eight of the tests I couldn't get running (and five of those are also broken in the github action). They seem orthogonal to this patch and fail the same way both before and after my patch.

Taking a peak at the automated GitHub action testing results in PR #33, it looks like that PR is also having the same five failures.

mentalisttraceur:
Since `tests/test_micropip.py` wasn't already using `hypothesis`, and since it would take much longer, I resisted the temptation to pull it in and significant inflate the test code with Hypothesis strategies for generating creative variations of wheel versions and tags.

Maybe that's a PR for a later time!

micropip/_micropip.py
```diff
@@ -89,6 +89,12 @@ def from_url(url: str) -> "WheelInfo":
...
 
+    def best_compatible_tag_index(self) -> int | None:
+        for index, tag in enumerate(sys_tags()):
+            if tag in self.tags:
+                return index
+        return None
+
     def is_compatible(self):
         if self.filename.endswith("py3-none-any.whl"):
             return True
```
rth, line 100: Maybe replace the section after this by,
```
return self.best_compatible_tag_index() is not None
```
to avoid having similar duplicate code paths.

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for adding mock packages, for use where something is a dependency and you don't need it, or you need only a limited subset of the package. This is done using `micropip.add_mock_package`, `micropip.remove_mock_package` and `micropip.list_mock_packages`. Packages installed like this will be skipped by dependency resolution when you later install real packages.
 
+### Fixed
+
+- Prioritization of wheels by Python version in their tags: for example,
+  if a package has two pure Python wheels, one tagged `py30` and
+  another tagged `py35`, the `py35` wheel will now always get installed.
+
```
rth, line 18:
```suggestion
  [#34](https://github.com/pyodide/pytest-pyodide/pull/34)

```

CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for adding mock packages, for use where something is a dependency and you don't need it, or you need only a limited subset of the package. This is done using `micropip.add_mock_package`, `micropip.remove_mock_package` and `micropip.list_mock_packages`. Packages installed like this will be skipped by dependency resolution when you later install real packages.
 
+### Fixed
+
+- Prioritization of wheels by Python version in their tags: for example,
```
rth, line 15:
```suggestion
- When multiple compatible builds for a package exist, the best build
  is now installed, as determined by the order of tags in  [`packaging.tags.sys_tags`](https://packaging.pypa.io/en/latest/tags.html#packaging.tags.sys_tags). For example,
```
as it's more general than Python version. It should also address pure python wheels vs compiled sources I think for instance.

micropip/_micropip.py
```diff
@@ -89,6 +89,12 @@ def from_url(url: str) -> "WheelInfo":
             parsed_url=parsed_url,
         )
 
+    def best_compatible_tag_index(self) -> int | None:
```
rth, line 92: Please add a docstring at least with a 1 line description

rth (commented):
Thanks a lot @mentalisttraceur for the fix and for writing tests!

Yes, we should definitely add contributing docs.

A couple of minor comments otherwise LGTM. Could you please merge upstream/main in the CI issues were fixed in https://github.com/pyodide/micropip/pull/35

micropip/_micropip.py
hoodmane: They are all supposed to have docstrings but sometimes I swindle Roman out of them.

PR #35, opened 2022-12-10
Pin to packaging 21.3 to fix CI

rth:
CI seems to be failing unrelated to different PRs (but it's green on master). Trying to reproduce it.


rth:
The CI issues were due to a new packaging release. For now, pinning to the previous good version to fix CI before deciding how we deal with the upgrade https://github.com/pyodide/micropip/issues/36

PR #40, opened 2022-12-12
Fix deployment CI

rth:
The deployment CI failed, so I ended up uploading the wheels manually (to avoid doing a post-release version tag).

Meanwhile, this is a fix for the deployment CI.


ryanking13:
Thanks!

PR #41, opened 2022-12-13
Fix installing underscore-prefixed unvendored stdlib

ryanking13:
`micropip.install("_hashlib")` was failing with:

```py
packaging.requirements.InvalidRequirement: Parse error at "'_hashlib'": Expected W:(0-9A-Za-z)
```

as it is not a valid package name. But we actually want to install it as it is unvendored stdlib.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        pyodide-version: [0.21.3, 0.22.0a3]
+        pyodide-version: [0.22.0a3]
         test-config: [
           {runner: selenium, runtime: chrome, runtime-version: latest },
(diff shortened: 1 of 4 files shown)


rth:
Maybe it would be better to rename `_hashlib` to `hashlib` and avoid leading underscore ? I mean if it's an invalid package name we should not use it.

ryanking13:
> Maybe it would be better to rename `_hashlib` to `hashlib` and avoid leading underscore ? I mean if it's an invalid package name we should not use it.

Okay, I'll try that instead.

> Also it looks like we have no tests in Pyodide for unvendored stlib loading? It's not the first time that one of them doesn't load.

Right, good point...

ryanking13:
Closing as we changed the name of _hashlib: https://github.com/pyodide/pyodide/pull/3349

PR #44, opened 2023-01-18
Apply dos2unix on everything

hoodmane:
The docs folder seems to have windows style line endings. This converts them to unix line endings.


rth:
Thanks! It's interesting that flake8 in CI started to fail, but it's probably unrelated.

PR #52, opened 2023-01-27
Use Python 3.11

ryanking13:


.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
-extend-ignore = E203, E402, E501, E731, E741, B950, W503, B028
+extend-ignore = E203, E402, E501, E731, E741, B950, W503, B028, B907
```
ryanking13, line 2: ["Rename B028 to B907, making it optional/opinionated."](https://github.com/PyCQA/flake8-bugbear#23117)

pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = [
 description = "A lightweight Python package installer for the web "
 readme = "README.md"
 license = { file="LICENSE" }
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
rth, line 9: Not sure about this though. It costs us nothing to keep to support 3.10 for now. Ideally tested in CI but even without testing. So that people could use future releases of micropip with the current release of pyodide.

If we support Pyodide 0.21.3, we need to support the corresponding Python version.

rth (approved):
LGTM, aside from one comment, thanks!

pyproject.toml
ryanking13: > So that people could use future releases of micropip with the current release of pyodide. If we support Pyodide 0.21.3, we need to support the corresponding Python version.

Yes, that makes sense. I'll remove that part.

PR #54, opened 2023-03-04
Skip invalid versions on PyPI

rth:
With this change, micropip would skip invalid versions (those that cannot be parsed with [packaging.version.Version](https://packaging.pypa.io/en/stable/version.html#version-handling)) when looking at available versions on PyPi.

Otherwise, with lastest `packaging` the current situation is to error, when there is even one invalid version in the list as was the case with pytz in 2004 in https://github.com/pyodide/pyodide/pull/3605#issuecomment-1442612037


ryanking13 (approved):
Thanks @rth!

Issue #51, opened 2023-01-26
Ability to reinstall a different version of a package


rth:
Currently if a package is already installed, we can't re-install a different version,
```py
>>> import micropip
>>> await micropip.install('packaging==23.0')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 573, in install
    await transaction.gather_requirements(requirements)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 333, in gather_requirements
    await gather(*requirement_promises)
  File "/lib/python3.10/asyncio/futures.py", line 284, in __await__
    yield self  # This tells Task to wait for completion.
  File "/lib/python3.10/asyncio/tasks.py", line 304, in __wakeup
    future.result()
  File "/lib/python3.10/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/lib/python3.10/asyncio/tasks.py", line 232, in __step
    result = coro.send(None)
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 340, in add_requirement
    return await self.add_requirement_inner(Requirement(req))
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 418, in add_requirement_inner
    if self.check_version_satisfied(req):
  File "/lib/python3.10/site-packages/micropip/_micropip.py", line 364, in check_version_satisfied
    raise ValueError(
ValueError: Requested 'packaging==23.0', but packaging==21.3 is already installed
```

On one side it's probably not ideal to do that, particularly given that micropip itself depends on packaging in this case. Still it feels like an arbitrary limitation. IMO in this case it would be better to uninstall the previous package version and install the new version.

ryanking13:
Yeah, sometimes we want to install a newer or old version during runtime, so I'm +1 for having the option to reinstall. A possible issue is that how we should deal with a package that is already imported.

rth:
> A possible issue is that how we should deal with a package that is already imported.

It's the same problem as installing things in a Jupyter notebook. IMO users would still need to explicitly either calling `importlib.reload` or using an auto-reload plugin like in Jupyter.

akshayka:
We're running into this issue in the marimo playground. The playground comes packaged with `narwhals 1.10.0`, but plotly currently requires a newer version of narwhals.

I noticed that #64 has stalled, due to concerns about the need to reload modules.

Is there a suggested workaround?

ryanking13:
[...]

PR #55, opened 2023-03-10
ENH Add micropip.uninstall()

ryanking13:
Related: #51 

Add `micropip.uninstall()` method, which uninstalls packages that are installed.

I used a quite simple approach for this compared to `pip uninstall`, since we only support installing wheels, and wheel metadata contain a list of files in the archive, so I just iterated through the list and removed them.

I didn't add a rollback feature if uninstall fails in this PR. I'm not sure it is worth it, but probably we can add it as a follow up.

- [x] changelog


ryanking13:
> What would happen if we uninstall a package installed via loadPackage? If it also uninstalls it, we need to be sure loadPackage would be aware of it.

I think removing keys in `loadedPackages` does the job. Will add a test.

ryanking13:
@rth Thanks for the review! I think I addressed all of your comments.

I changed the code a bit after your review (simplified how files are retrieved, and added tests), so I would appreciate it if you can review this one more time.

micropip/utils.py
```diff
@@ -0,0 +1,49 @@
+from importlib.metadata import Distribution
+from pathlib import Path
+
+
+def get_dist_info(dist: Distribution) -> Path:
```
rth, line 5: Can you either make these functions private (starting with _) or make the whole module private?

rth (approved):
One minor comment, otherwise LGTM. Thanks!

PR #60, opened 2023-04-14
Add verbose option to install and uninstall

ryanking13:
This adds `verbose` parameter to `micropip.install` and `micropip.uninstall`.

Setting verbose=True will print the installation/uninstallation process.

image

A known issue is that it does not show dependencies of packages in repodata.json (as it is handled by `loadPackage`)

image

- [x] changelog

### DIFF
--- a/micropip/_utils.py
+++ b/micropip/_utils.py
@@ -47,9 +47,3 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
     files_to_remove.update(metadata_files)
 
     return files_to_remove
-
-
-def log(message: str, verbose: bool = False):
-    # TODO: use logging instead of print
-    if verbose:
-        print(message)
--- a/micropip/_utils.py
+++ b/micropip/_utils.py
@@ -47,3 +47,9 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
     files_to_remove.update(metadata_files)
 
     return files_to_remove
+
+
+def log(message: str, verbose: bool = False):
+    # TODO: use logging instead of print
+    if verbose:
+        print(message)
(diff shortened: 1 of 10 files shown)


micropip/_utils.py
```diff
@@ -47,3 +47,9 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
     files_to_remove.update(metadata_files)
 
     return files_to_remove
+
+
+def log(message: str, verbose: bool = False):
+    # TODO: use logging instead of print
+    if verbose:
+        print(message)
```
rth, line 55: This is never used?

micropip/transaction.py
```diff
@@ -355,24 +363,31 @@ def eval_marker(e: dict[str, str]) -> bool:
...
 
-        await self.add_wheel(wheel, req.extras)
+        await self.add_wheel(wheel, req.extras, specifier=str(req.specifier))
 
     async def add_wheel(
         self,
         wheel: WheelInfo,
         extras: set[str],
+        *,
+        specifier: str = "",
```
rth, line 380: Not very clear to me out of context what the specifier is supposed to be. Can you maybe add a docstring for it (I imagine it's documented in req.specifier)?

rth (approved):
Great thanks a lot @ryanking13 ! Overall LGTM.

micropip/_utils.py
ryanking13: Right, I forgot to remove this :)

micropip/transaction.py
ryanking13: Added docstring for that method. Actually that parameter is used only for logging so I guess we need to remove that in the future.

PR #63, opened 2023-04-24
Make error message more helpful

jasongrout:
This adds an `await` to help the user use the suggested code.


ryanking13 (approved):
Thanks!

PR #64, opened 2023-04-25
ENH Add support for reinstalling packages

### LINKED ISSUE #51: Ability to reinstall a different version of a package

ryanking13:
Resolve #51

~~This PR changes `micropip.install` behavior to remove previously installed packages and reinstall the new version. In addition, it adds the `force_reinstall` parameter to `micropip.install`, allowing it to force a reinstall even if it already meets the requirements.~~

_Edit:_

This PR adds `reinstall` parameter to micropip.install which allows reinstalling packages that are already installed.

- [x] changelog
- [x] refactorings


ryanking13:
@rth I just noticed that you added `sprint` label in #51. Maybe there is a sprint happening in PyconUS? Sorry if I hijacked someones work.

rth:
Yes, there is a sprint in PyConUS so I tagged a few issues people have more to choose from. No worries if you took it :)

hoodmane:
Thanks @ryanking13!

> remove previously installed packages and reinstall the new version

I don't think this makes sense as a default. In my opinion it should not reinstall anything unless it is specifically requested.

ryanking13:
> I don't think this makes sense as a default. In my opinion it should not reinstall anything unless it is specifically requested.

Maybe my description was not clear. It uninstalls only if the requested version is incompatible with the installed version. Previously in such cases,  micropip failed with:

```
ValueError: Requested 'packaging==22.0', but packaging==23.0 is already installed
```

hoodmane (commented):
I still think failing is better by default than reinstall. I think people really mostly don't want to reinstall packages that they've already imported but they might not realize they don't want to. Maybe having an allow reinstall flag would be good.

At the wasm summit they were asking if they could test packaging against wasm and I was like sure easy but then the command line runner had some weird crash...

rth:
> If ``True``, reinstall all packages even if they are already up-to-date.

Yeah, I'm also not sure about it. As a user, in the example above I would expect the `install` command to only change the packages that need changing and that I requested. But I guess it's more complex in terms of dependency resolution.  This feels like re-installing a venv, which is good but strange to do in Python when not managing venv.

If we take the case where we install some packages. Some are already installed with different versions. If the version is compatible we do nothing, if it isn't we uninstall them, and install the requested version (and leave all other packages unchanged). That's more or less what pip does (or used to do) by default. Can't we do that?

ryanking13:
> I still think failing is better by default than reinstall. I think people really mostly don't want to reinstall packages that they've already imported but they might not realize they don't want to. Maybe having an allow reinstall flag would be good.

Hmm, yes. Unlike other package managers micropip is a runtime package manager so the situation is more complex that some packages might be already imported... and we don't show any logs (unless verbose option is enabled in #60),  so people will not even notice that some packages are reinstalled.

> If we take the case where we install some packages. Some are already installed with different versions. If the version is compatible we do nothing, if it isn't we uninstall them, and install the requested version (and leave all other packages unchanged). That's more or less what pip does (or used to do) by default. Can't we do that?

Yes, that is the current expected bahavior but I guess Hood does not prefer uninstalling packages without explicit flag?

ryanking13:
Then how about:

1. (default) `micropip.install(reinstall=False)` will show error messages if there is an installed version that is incompatible.

2. `micropip.install(reinstall=True)` will reinstall packages that are incompatible / keep packages that are compatible (same as pip).

ryanking13:
> Well if it's for matching pip then I could sort of see it. But the issue is just what happens if the package is already imported...

Well jupyter shares the same problem and users would need to call `importlib.reload` if they have already imported a package. One difference is that pip is verbose by default while micropip is silent so users might not notice that they have reinstalled a package.

ryanking13:
Closing in favor of #206

Issue #62, opened 2023-04-17
Plans for using Simple Repository API


ryanking13:
I've recently started working on changing micropip to use the simple API ([PEP 503](https://peps.python.org/pep-0503/), [PEP 691](https://peps.python.org/pep-0691/)), instead of [legacy JSON API](https://github.com/pyodide/micropip/blob/926caa78f75d7382ae88960029f60a9d318b4e25/micropip/transaction.py#L450). I think changing to simple API will help people to create alternative package registries following the standard, and we can also benefit from other PyPA tools.

- [x] Add CORS headers to https://pypi.org/simple (https://github.com/pypi/warehouse/pull/13222)
- [ ] Change micropip's default endpoint to https://pypi.org/simple
  - [brettcannon/mousebender](https://github.com/brettcannon/mousebender) is a useful reference for parsing simple API.
- [ ] Use [devpi](https://github.com/devpi/devpi) or [pypi/warehouse](https://github.com/pypi/warehouse) to host wheels for test
  - replace mock functions with a private PyPI server.
  - alternatives: [pypiserver](https://github.com/pypiserver/pypiserver), [pypa/bendersnatch](https://github.com/pypa/bandersnatch)
    - pypiserver is simple to use, but it is not actively maintained, and it doesn't seem to support PEP 691.
    - pypa/bendersnatch is for mirroring and doesn't seem to support uploading local wheels.
- [x] Support alternative registries that use simple API (#7)
- [ ] PEP658

ryanking13:
For now, this is blocked by: https://github.com/pypi/warehouse/issues/12214

rth:
Thanks a lot for looking into this @ryanking13 !

> Use [devpi](https://github.com/devpi/devpi) or [pypi/warehouse](https://github.com/pypi/warehouse) to host wheels for test

See also discussion in https://github.com/pyodide/pyodide/issues/3049 though I haven't really made much progress on it. Aside from determining that warehouse is not suitable for this purpose according to its devs, as it's the code for PyPI and too complex to use for other applications.

rth:
> some stale responses does not contain CORS headers added in 
> Change micropip's default endpoint to [pypi.org/simple](https://pypi.org/simple)

Would it be too much work to support both? I mean this would be great to support other hosting solutions that could use the Simple API even independently from PyPI. For instance,
 - https://pypi.anaconda.org/beeware/simple/regex/ 
 - https://jfrog.com/artifactory/
 
to name a new.

[...]

PR #65, opened 2023-04-26
Switch to Simple API

ryanking13:
This PR works, but it is for now blocked by: https://github.com/pyodide/micropip/issues/62#issuecomment-1515731076

- Change PyPI endpoint from JSON API (https://pypi.org/pypi/{pkgname}/json) to Simple API (https://pypi.org/simple)
- Vendored part of https://github.com/brettcannon/mousebender, which handles Simple API response.


rth:
So what do you think we should about this @ryanking13 ? Do you think we could maybe add Sample API support while keeping current behavior for PyPI for now, or that would be too complex? I mean it does sound like there could be some fastly cache related improvements https://github.com/pypi/warehouse/issues/12214#issuecomment-1598870736 but I don't really understand how blocking this situation currently is.

ryanking13:
@rth Nice timing. I was about to start working on this PR again.

> Do you think we could maybe add Sample API support while keeping current behavior for PyPI for now, or that would be too complex?

I think we can go with supporting both Simple and legacy JSON API, (hopefully) it won't be very complex. 

> I mean it does sound like there could be some fastly cache related improvements https://github.com/pypi/warehouse/issues/12214#issuecomment-1598870736 but I don't really understand how blocking this situation currently is.

I think now the cache issue is okay. I mean fastly issue is still not resolved, but at least now all Simple JSON response in PyPI contains the CORS header correctly, so we can use it in micropip.

ryanking13:
Continued on #75.

PR #66, opened 2023-04-26
Run tests against Pyodide 0.23.2

ryanking13:


ryanking13:
I seems like there was a bug in `micropip.add_mock_package` in Pyodide>=0.23.0.

.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        pyodide-version: [0.21.3, 0.22.1]
+        pyodide-version: [0.21.3, 0.22.1, 0.23.1]
```
hoodmane, line 21: Maybe 0.23.2 now...

micropip/_commands/mock_package.py
```diff
@@ -81,85 +80,99 @@ def init_fn(module):
...
-        file_list.append(metadata_file)
-        file_list.append(installer_file)
-        with open(metadata_file, "w") as mf:
-            mf.write(METADATA)
-        for n, s in modules.items():
-            if s is None:
-                s = ""
-            s = dedent(s)
-            path_parts = n.split(".")
+        site_packages = Path(site.getsitepackages()[0])
```
hoodmane, line 101: I am not a fan of this very long conditional, especially since the else is very short and it's basically the end of the function. Ideally we should change this to:
```
   if not persistent:
      # short not persistent code
      return
   # longer persistent code not indented
```
or it could even be factored into a separate function.

I understand this isn't really your fault though @ryanking13. If you want we can clean it up later.

hoodmane (approved):

micropip/_commands/mock_package.py
ryanking13: Yeah, I'll do some cleanup later.

PR #67, opened 2023-05-02
Add devpi docker image

ryanking13:
This PR adds [devpi](https://github.com/devpi/devpi) docker image, which is a private PyPI server that can be used for testing micropip.

devpi is not integrated with micropip tests yet, since micropip does not support alternative registries yet. I will open follow-up PRs.

This PR includes:

- Dockerfile / docker-compose.yml that builds and runs devpi server.
- `pytest-docker` package that integrates devpi docker with pytest
-  Some basic tests that check devpi
- A plugin that adds PyPI JSON API support in devpi


devpi/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+# Adapted from: https://github.com/LordGaav/docker-devpi
+
+function generate_password() {
+    # We disable exit on error because we close the pipe
+    # when we have enough characters, which results in a
+    # non-zero exit status
+    set +e
+    tr -cd '[:alnum:]' < /dev/urandom | fold -w30 | head -n1 | tr -cd '[:alnum:]'
+    set -e
+}
```
rth, line 14: Maybe we could set some fixed password to make it simpler? Of in a CI env variable if you prefer to keep it a secret. We don't really care about it being random for testing.

devpi/entrypoint.sh
```diff
@@ -0,0 +1,75 @@
...
+    kill -SIGTERM "$_PID"
+}
+
+if [ "${1:-}" == "bash" ]; then
+    exec "$@"
+fi
+
+DEVPI_ROOT_PASSWORD="${DEVPI_ROOT_PASSWORD:-}"
+if [ -f "$DEVPI_SERVER_ROOT/.root_password" ]; then
+    DEVPI_ROOT_PASSWORD=$(cat "$DEVPI_SERVER_ROOT/.root_password")
```
rth, line 28: All of this can be removed in our case probably. This script can just expect `DEVPI_ROOT_PASSWORD` env variable to be set?.

devpi/nginx.conf
```diff
@@ -0,0 +1,20 @@
...
+    listen       13141;
+    server_name  localhost;
+location /
+    {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+    }
+location /health
+    {
+        return 200 '{"message": "healthy"}';
```
rth, line 12: What's the point of this endpoint, if it returns a constant anyway?

devpi/nginx.conf
```diff
@@ -0,0 +1,20 @@
...
+        add_header Content-Type application/json;
+    }
+location /devpi/
+    {
+        proxy_pass http://devpi:3141/;
+        add_header 'Access-Control-Allow-Origin' '*' always;
```
rth, line 18: Can we open an issue at devpi to see if they can optionally add CORS ?

devpi/plugins/devpi-json-info/devpi_json_info.py
```diff
@@ -0,0 +1,68 @@
...
+
+@view_config(
+    route_name="json_info",
+    accept="application/json",
+    request_method="GET",
+    renderer="json",
+)
+def json_info_view(context, request) -> dict[str, Any]:
+    # Override the base_url with the BASE_URL environment variable
+    # This is useful when running behind a reverse proxy
```
rth, line 32: Wait so the sha256 changes if run behind a reverse proxy? I would have guessed it shouldn't. 

Isn't the point to use stock devpi server?

tests/conftest.py
```diff
@@ -232,3 +233,50 @@ def mock_fetch(monkeypatch, mock_importlib):
...
     return result
+
+
+@pytest.fixture(scope="session")
+def docker_compose_file():
+    return str(Path(__file__).parent.parent / "devpi/docker-compose.yml")
+
+
+@pytest.fixture(scope="session")
+def devpi(docker_ip, docker_services):
```
rth, line 244: IMO we should be able to run the test suite without relying on docker: so if would be better if this was opt-in.

rth:
Thanks @ryanking13!  Some comments are above.

I see the value of having a real server for Python wheels but I think Docker (or installing this server) ideally shouldn't be required to run tests (or at least it should be optional) As far as I know pip doesn't do this for tests. The other alternative is to use this test server, record the response and save them to the file system then mock the response with it. It's a bit more work though.

Maybe @pradyunsg would have some guidance on the best way to test compliance with a server that exposes a PyPI-like JSON API?

ryanking13:
> I see the value of having a real server for Python wheels but I think Docker (or installing this server) ideally shouldn't be required to run tests (or at least it should be optional) As far as I know pip doesn't do this for tests. The other alternative is to use this test server, record the response and save them to the file system then mock the response with it. It's a bit more work though.

Yes, that is a good point. I was also a bit concerned about making docker mandatory to run tests. I like the idea of recording the responses from the test server. Basically, I wanted to remove our mock_fetch class, which is very hard to maintain I think.

ryanking13:
Closing as we added a way to run a mock server in #74.

PR #71, opened 2023-07-05
Add a common class to store JSON / Simple API response data

ryanking13:
This adds an intermediate class `ProjectInfo` (I can't think of a better name... open to other names) that parses both JSON and Simple API PyPI responses. micropip now takes this `ProjectInfo` class instead of a raw PyPI JSON response when searching for packages

This is a preparation to make micropip support both JSON and Simple API.

Related: #62

### DIFF
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -52,12 +52,8 @@ class ProjectInfo:
         # Filter out non PEP 440 compliant versions
         releases: dict[Version, list[Any]] = {}
         for version_str, fileinfo in releases_raw.items():
-            try:
-                version = Version(version_str)
-                if str(version) != version_str:
-                    continue
-
-            except InvalidVersion:
+            version, ok = _is_valid_pep440_version(version_str)
+            if not ok or not version:
                 continue
 
             # Skip empty releases
@@ -86,11 +82,12 @@ class ProjectInfo:
         # Group files by version
         releases: dict[Version, list[Any]] = defaultdict(list)
 
-        for version in versions:
-            if not _is_valid_pep440_version(version):
+        for version_str in versions:
+            version, ok = _is_valid_pep440_version(version_str)
+            if not ok or not version:
                 continue
 
-            releases[Version(version)] = []
+            releases[version] = []
 
         for file in data["files"]:
             filename = file["filename"]
@@ -157,15 +154,17 @@ class ProjectInfo:
         )
 
 
-def _is_valid_pep440_version(version_str: str) -> bool:
+def _is_valid_pep440_version(version_str: str) -> tuple[Version | None, bool]:
+    """
+    Check if the given string is a valid PEP 440 version.
+    Since parsing a version is expensive, we return the parsed version as well,
+    so that it can be reused if needed.
+    """
     try:
         version = Version(version_str)
-        if str(version) != version_str:
-            return False
-
-        return True
+        return version, True
     except InvalidVersion:
-        return False
+        return None, False
 
 
 def _fast_check_incompatibility(filename: str) -> bool:
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -250,27 +250,6 @@ async def test_install_pre(
 
 
 @pytest.mark.asyncio
-@pytest.mark.filterwarnings("ignore::Warning")
-@pytest.mark.parametrize("version_invalid", ["1.2.3-1", "2.3.1-post1", "3.2.1-pre1"])
-async def test_install_version_invalid_pep440(
-    mock_fetch: mock_fetch_cls,
-    version_invalid: str,
-) -> None:
-    # Micropip should skip package versions which do not follow PEP 440.
-    #
-    #     [N!]N(.N)*[{a|b|rc}N][.postN][.devN]
-    #
-
-    dummy = "dummy"
-    version_stable = "1.0.0"
-
-    mock_fetch.add_pkg_version(dummy, version_stable)
-    mock_fetch.add_pkg_version(dummy, version_invalid)
-    await micropip.install(dummy)
-    assert micropip.list()[dummy].version == version_stable
-
-
-@pytest.mark.asyncio
 async def test_fetch_wheel_fail(monkeypatch, wheel_base):
     pytest.importorskip("packaging")
     import micropip


micropip/package_index.py
```diff
@@ -0,0 +1,121 @@
...
+        for version_str, fileinfo in _releases.items():
+            try:
+                version = Version(version_str)
+            except InvalidVersion:
+                # Ignore non PEP 440 compliant versions
+                continue
+
+            for file in fileinfo:
+                filename = file["filename"]
+                if not filename.endswith(".whl"):
```
rth, line 60: Would it make sense to filter the compatible python versions / platform here? I mean keeping all those wheels that cannot be used is not very useful.

micropip/package_index.py
```diff
@@ -0,0 +1,121 @@
...
+            releases[version].append(
+                ProjectInfoFile(
+                    filename=filename,
+                    url=file["url"],
+                    version=version,
+                    # TODO: For now we always expect that the sha256 hash is available.
+                    # This is true for PyPI, but may not be true for other package indexes,
+                    # since it is not a hard requirement of PEP503.
+                    sha256=file["hashes"]["sha256"],
+                    size=file["size"] if "size" in file else None,
```
rth, line 114:
```suggestion
                    size=file.get('size'),
```

micropip/package_index.py
```diff
@@ -0,0 +1,121 @@
...
+                if not filename.endswith(".whl"):
+                    continue
+
+                releases[version].append(
+                    ProjectInfoFile(
+                        filename=filename,
+                        url=file["url"],
+                        version=version,
+                        sha256=file["digests"]["sha256"],
+                        size=file["size"] if "size" in file else None,
```
rth, line 69:
```suggestion
                        size=file.get("size"),
```

micropip/package_index.py
```diff
@@ -0,0 +1,121 @@
...
+
+        name = data["name"]
+        releases: defaultdict[Version, list[ProjectInfoFile]] = defaultdict(list)
+        for file in data["files"]:
+            filename = file["filename"]
+            if not filename.endswith(".whl"):
+                continue
+
+            try:
+                version = parse_wheel_filename(filename)[1]
```
rth, line 100: A bit unfortunate to parse the wheel filename twice once here, and once when filtering out by platform, but it's fast enough, so shouldn't matter.
```
In [2]: %timeit parse_wheel_filename('scikit_learn-1.3.0-cp311-cp311-manylinux_2_17_x86_64.manylinux2014_x86_64.whl')
6.06 µs ± 31.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

tests/test_package_index.py
```diff
@@ -0,0 +1,72 @@
...
+import pytest
+
+import micropip.package_index as package_index
+
+TEST_TEMPLATES_DIR = Path(__file__).parent / "test_data" / "pypi_response"
+
+
+@pytest.mark.parametrize("name", ["black", "pytest", "snowballstemmer"])
+def test_project_info_from_json(name):
+    test_file = TEST_TEMPLATES_DIR / f"{name}_json.json"
```
rth, line 13: Can you gzip the json with data, so that grepping the source code wouldn't produce all sorts of noise with text data?

tests/test_package_index.py
```diff
@@ -0,0 +1,72 @@
...
+    test_data = json.loads(test_file.read_text())
+
+    index = package_index.ProjectInfo.from_json_api(test_data)
+    assert index.name == name
+    assert index.releases
+
+    versions = list(index.releases.keys())
+    assert versions == sorted(versions)
+
```
rth, line 22: Maybe we could also assert the number of entries?

micropip/package_index.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import Any
+
+from packaging.utils import (
+    parse_wheel_filename,
+)
+from packaging.version import InvalidVersion, Version
+
+
+@dataclass
+class ProjectInfoFile:
```
rth, line 12: `WheelInfo` maybe? That's a similar container to https://github.com/pyodide/pyodide/blob/main/pyodide-build/pyodide_build/mkpkg.py#L23 BTW might we worth mentioning it in a comment.

rth (commented):
Thanks @ryanking13 that would be very useful indeed! 

A couple of comments. Ping me when it's ready for a final review.

micropip/package_index.py
ryanking13: Yes, I mentioned that in the comment above. One concern is that usually, we don't need to iterate through the whole list of versions because, when we do `micropip.install` it will often only look at the latest version.

I'll check if the filtering process is fast enough. If not I'll try using a generator.

tests/test_package_index.py
ryanking13: Good point! Will do.

micropip/package_index.py
ryanking13: We are already using that name here.

https://github.com/pyodide/micropip/blob/501a6910e79181b5a9ccfefa2ec88728a99e04d6/micropip/transaction.py#L37

Maybe we can refactor these two classes in a follow-up, by the way. They contain similar attributes.

rth: Ahh yes. Maybe make this `BaseWheelInfo` and inherit from it there, since it's a subset of those fields?

ryanking13: > We can also do it later, but someone would need to remember it. I probably wont :)

This (Simple API and related stuff) is my high-priority work for now, so I'll remember. I'll add some comments so I won't forget :)

micropip/_utils.py
```diff
@@ -54,6 +56,11 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
     return files_to_remove
 
 
+@functools.lru_cache
```
hoodmane, line 59:
```suggestion
@functools.cache
```

ryanking13:
@rth Okay, I think this is ready to be reviewed.

I cached `sys_tags` and `parse_wheel_filename` since those functions were bottlenecks when I checked.

In terms of performance, `from_simple_api` is slower than `from_json_api`. This is because simple API does not have a version key in the response so we need to parse the wheel filename for every file. I added a few checks to optimize the speed but it is still around 2 times slower.

For example, this is a result when I tested with `numpy`, which has ~2000 files in the index:

```python
from micropip.package_index import ProjectInfo
from pathlib import Path
from gzip import decompress
import json
import timeit

numpy_json = json.loads(decompress(Path("tests/test_data/pypi_response/numpy_json.json.gz").read_bytes()))
numpy_simple = json.loads(decompress(Path("tests/test_data/pypi_response/numpy_simple.json.gz").read_bytes()))
num_runs = 100

def run_json():
    ProjectInfo.from_json_api(numpy_json)

def run_simple():
    ProjectInfo.from_simple_api(numpy_simple)

print("Total number of files:", len(numpy_simple["files"]))
print(f"Parsing JSON API for {num_runs} times:", timeit.timeit(run_json, number=num_runs), "seconds")
print(f"Parsing Simple API for {num_runs} times:", timeit.timeit(run_simple, number=num_runs), "seconds")

```

```
Total number of files: 2185
Parsing JSON API for 100 times: 0.11372180000762455 seconds
Parsing Simple API for 100 times: 0.243828200007556 seconds
```

micropip/_utils.py
```diff
@@ -47,3 +55,129 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
...
+
+    try:
+        tags = parse_tags(filename)
+    except (InvalidVersion, InvalidWheelFilename):
+        return False
+
+    return best_compatible_tag_index(tags) is not None
+
+
+def check_compatible(filename: str) -> None:
```
rth, line 127: Nice refactoring of these functions!

micropip/package_index.py
```diff
@@ -0,0 +1,188 @@
...
+    def _compatible_only(
+        name: str, releases: dict[Version, list[dict[str, Any]]]
+    ) -> "ProjectInfo":
+        def _compatible_wheels(
+            files: list[dict[str, Any]], version: Version
+        ) -> Generator[ProjectInfoFile, None, None]:
+            """
+            Return a generator of wheels compatible with the current platform.
+            Checking compatibility takes a bit of time, so we use a generator to avoid doing it if not needed.
+            """
```
rth, line 121: The docstring should be on the outer function?

micropip/package_index.py
```diff
@@ -0,0 +1,188 @@
...
+                if not compatible:
+                    continue
+
+                # JSON API has a "digests" key, while Simple API has a "hashes" key.
+                hashes = file["digests"] if "digests" in file else file["hashes"]
+
+                # TODO: For now we expect that the sha256 hash is always available.
+                # This is true for PyPI, but may not be true for other package indexes,
+                # since it is not a hard requirement of PEP503.
+                sha256 = hashes["sha256"]
```
rth, line 137: As long as you remember to fix it later.
Otherwise we could put,
```py
hashes =   file["digests"] if "digests" in file else file.get("hashes", {})
sha256 = hashes.get('sha256')
```
already now.

micropip/package_index.py
```diff
@@ -0,0 +1,188 @@
...
+                version = parse_version(filename)
+            except (InvalidVersion, InvalidWheelFilename):
+                continue
+
+            releases[version].append(file)
+
+        return ProjectInfo._compatible_only(name, releases)
+
+    @staticmethod
+    def _compatible_only(
```
rth, line 112: This should probably be more a classmethod since it returns an instance of this class?

rth (approved):
Nice refactoring @ryanking13 ! A couple of minor comments, otherwise LGTM.

micropip/package_index.py
```diff
@@ -0,0 +1,188 @@
...
+        name: str = data.get("info", {}).get("name", "UNKNOWN")
+        releases_raw: dict[str, list[Any]] = data["releases"]
+
+        # Filter out non PEP 440 compliant versions
+        releases: dict[Version, list[Any]] = {}
+        for version_str, fileinfo in releases_raw.items():
+            try:
+                version = Version(version_str)
+                if str(version) != version_str:
+                    continue
```
rth, line 58: Do we need to keep this? As long as there is a version it's fine no? And str(version) has a non negligible runtime.

micropip/package_index.py
```diff
@@ -0,0 +1,188 @@
...
+                yield ProjectInfoFile(
+                    filename=filename,
+                    url=file["url"],
+                    version=version,
+                    sha256=sha256,
+                    size=file.get("size"),
+                )
+
+        releases_compatible = {
+            version: _compatible_wheels(files, version)
```
rth, line 148: If we want to go faster (no need to do it now). We can probably also subclass Version to something that has a frozen __str__ and __hash__ so it doesn't need to be re-computed at each dict comparison..

micropip/_utils.py
```diff
@@ -47,3 +55,129 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
...
+    version = f"{version_info.major}{version_info.minor}"
+    abis = ["abi3", f"cp{version}"]
+    for tag in tags:
+        if tag.abi in abis:
+            abi_incompatible = False
+        break
+    if abi_incompatible:
+        abis_string = ",".join({tag.abi for tag in tags})
+        raise ValueError(
+            f"Wheel abi '{abis_string}' is not supported. Supported abis are 'abi3' and 'cp{version}'."
```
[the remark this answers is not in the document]
hoodmane: If we are planning to break ABI with every Python version, then abi3 is a bit pointless. I think we can support it and just check that the interpreter version matches exactly. So it's supposed to be that `cp311-abi3` works with all Pythons <= 311 but instead we check that the interpreter is exactly `cp311`. For tools like cibuildwheel that generate wheels, there's no reason to support abi3. But if someone uses a tool like maturin to generate an abi3 wheel we should be able to consume it.

ryanking13:
> Thanks it's good to know! In what remains it looks like Version.__str__ and Version.__hash__ (e.g. when using as a dict key) is also non negligible (run with pyinstrument),

Right, converting string <--> Version takes quite a time... I'm not sure how many people will have performance issues with this (especially given that network latency is a much larger part of it), but it's definitely something we can optimize for. By the way, pyinstrument is cool. Thanks for introducing an awesome tool :)

micropip/_utils.py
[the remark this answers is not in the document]
ryanking13: Looks like we have some packages that has `abi3` abi tags in Pyodide distribution.

image

micropip/package_index.py
ryanking13: I just noticed that micropip was already handling cases when checksum is not available.

micropip/package_index.py
ryanking13: Good point. I think it is okay to remove it now.

That check was introduced at https://github.com/pyodide/pyodide/issues/2752, but I think it cannot happen on wheel files (unless someone modifies the version manually).

micropip/package_index.py
ryanking13: I opened an issue (https://github.com/pyodide/micropip/issues/73) so that we can do it as a follow-up. I also think it is would be a good issue for new contributors if there is other sprint / tutorial events.

ryanking13:
Thanks for the reviews!

PR #74, opened 2023-07-17
ENH Support alternative index urls

### LINKED ISSUE #7: Support alternative registries in micropip
We have discussed this from time to time, but I can't find an issue about it.

It would be good to support alternative registries in micropip. We already support installation from custom URLs, however, a registry that supports the wheelhouse JSON protocol (compatible wirth PyPi) would be much faster and more convenient to use for dependency resolution.

Maybe this would be a global setting,
```py
micropip.set_registry('"https://some-registory.com/endpoint/")
```
with the default being `https://pypi.org/pypi/`. The only place it's used [is here](https://github.com/pyodide/pyodide/blob/main/packages/micropip/src/micropip/_micropip.py#L85).

Alternatively, we could support a list of channels ordered by priority. That would be maybe better but a bit more difficult to implement correctly.

ryanking13:
Resolve #7

This PR allows users to specify an index URL from which micropip can download packages. This allows users to install packages from a custom package index that provides PyPI-compatible APIs.

This PR covers both methods: modifying the index URL globally (`micropip.set_index_urls`) and passing the index URL as a parameter to `micropip.install`. This allows service operators like pyscript / jupyterlite to add their own package indexes as a default setting, while also allowing users to choose and use multiple indexes when needed.

- [x] changelog


micropip/package_index.py
```diff
@@ -182,3 +190,59 @@ def _fast_check_incompatibility(filename: str) -> bool:
...
+
+    if "package_name" not in fields:
+        raise ValueError(
+            f"Invalid index URL: {url!r}. "
+            "Please make sure it contains the placeholder {package_name}."
+        )
+
+
+async def search_packages(
+    pkg: str,
```
rth, line 206: Should it be `search_package` if it searches just 1 package? Also, it might be more user friendly to name the first argument `name`. 

It's not "search" technically though. Search is returning top k results for a query, possibly with some pre-processing or fuzzing of the query. While this return a single response with which is the exact match. Maybe `query_package`? But I'm open to other name suggestions.

micropip/_commands/install.py
```diff
@@ -86,6 +87,13 @@ async def install(
         If ``True``, include pre-release and development versions. By default,
         micropip only finds stable versions.
 
+    index_urls :
+
+        A list of URLs or a single URL to use as the package index.
+        By default, micropip uses the PyPI. The URLs should contain
```
rth, line 93: It would be good indicating what's the actual URL value when None is used (`"https://pypi.org/pypi/{package_name}/json"`). As in if I want to add another index URL in addition to the PyPI one, how do I do that.

Also,
 - Would it be good to indicate that these URLs need to expose a simple JSON API, aside from PyPI? 
 -  worth adding a sentence about the ordering of this list, indicating that the package is first searched in the first index, then if not found in the second and so on.

micropip/package_index.py
```diff
@@ -182,3 +190,59 @@ def _fast_check_incompatibility(filename: str) -> bool:
...
+        try:
+            metadata = await fetch_string(url, fetch_kwargs)
+        except OSError:
+            continue
+
+        return ProjectInfo.from_json_api(json.loads(metadata))
+    else:
+        raise ValueError(
+            f"Can't fetch metadata for '{pkg}'."
+            "Please make sure you have entered a correct package name."
```
rth, line 247: and correctly specified `index_urls` (if you changed them).

rth (commented):
Very nice! A few comments otherwise LGTM. Thanks @ryanking13 !

micropip/package_index.py
ryanking13: Thanks for the suggestion! I wasn't aware of the difference between search and query :) I will go with `query_package`.

micropip/_commands/install.py
ryanking13: > Would it be good to indicate that these URLs need to expose a simple JSON API, aside from PyPI?

Right, we should mention exactly what kind of APIs we support.

ryanking13:
I removed the `<package_name>` placeholder limitation so that index URLs for Simple APIs (e.g. `https://pypi.org/simple` can be passed without modification.

PR #75, opened 2023-07-19
Add support for Simple Repository API (PEP 503 / PEP 691)

ryanking13:
This adds support for Simple Repository API including [PEP 691](https://peps.python.org/pep-0691/) and [PEP 503](https://peps.python.org/pep-0503/).
This PR also changes the default index URL from `https://pypi.org/pypi/{package_name}/json` to `https://pypi.org/simple`.

I added some tests that send requests to remote package indexes like PyPI, so we can ensure that micropip "works" with famous package indexes. Those tests don't run by default in CI, but one can run them manually by adding `--remote-remote-index-tests` flag or by clicking the button in GHA.

- [ ] changelog


micropip/package_index.py
```diff
@@ -2,17 +2,19 @@
...
 from packaging.utils import InvalidWheelFilename
 from packaging.version import InvalidVersion, Version
 
-from ._compat import fetch_string
+from ._compat import fetch_string_and_headers
 from ._utils import is_package_compatible, parse_version
+from .externals.mousebender.simple import from_project_details_html
 
-DEFAULT_INDEX_URLS = ["https://pypi.org/pypi/{package_name}/json"]
+DEFAULT_INDEX_URLS = ["https://pypi.org/simple"]
```
rth, line 17: Can you confirm we are never actually downloading the top level index here, because it's >3MB compressed in the case of PyPI.

rth (approved):
Very nice, thanks @ryanking13 ! I have only one minor comment otherwise LGTM. 

I think once this is merged we should do a micropip release with it. Unless there were other PRs you wanted to do in this area.

micropip/package_index.py
ryanking13: Yes, I can confirm that we are not downloading the top level index, unless someone changes the index URL to `https://pypi.org` and do `micropip.install("simple")` :)

ryanking13:
> I think once this is merged we should do a micropip release with it. Unless there were other PRs you wanted to do in this area.

Yes, I'll make a release so we can include it in Pyodide 0.24.

PR #79, opened 2023-09-01
Options to fix dependencies of manually loaded packages

joemarshall:
Right now, if you import packages from wheel files without automatically getting dependencies, `micropip.freeze` won't correctly show dependencies for the packages.

This patch fixes up dependencies for manually loaded wheels, adding a method `micropip.check_package_dependencies` to check and fix dependencies for a package.

It also uses the same function to make `micropip.freeze` work even if you just manually load a wheel and all dependencies. 

Still not sure whether this would be better in pyodide-lock, but I think given `micropip.freeze` exists, this patch that makes it work in this case seems reasonably sensible.


micropip/_commands/freeze.py
```diff
@@ -37,7 +37,13 @@ def freeze() -> str:
             depends = json.loads(requires)
         else:
             depends = []
-
+            if dist.requires is not None and len(dist.requires) > 0:
```
hoodmane, line 40: How about:
```py
requires = dist.read_text("PYODIDE_REQUIRES")
if not requires:
   check_package_dependencies(name, fix_deps=True)
   requires = dist.read_text("PYODIDE_REQUIRES")
if requires:
   depends = json.loads(requires)
else:
   depends = []
```

micropip/_commands/check_package_dependencies.py
```diff
@@ -0,0 +1,105 @@
...
+    # can't do anything with it.
+    assert url is not None
+
+    # Get current list of pyodide requirements
+    requires = dist.read_text("PYODIDE_REQUIRES")
+
+    if requires:
+        depends = json.loads(requires)
+    else:
+        depends = []
```
hoodmane, line 60: Since this is used here and also in `freeze.py` can we factor this out into a utility method?

micropip/_commands/check_package_dependencies.py
```diff
@@ -0,0 +1,105 @@
...
+
+    missing = []
+    all_missing = {}
+
+    package_requires = dist.requires
+    if package_requires is None:
+        # no dependencies - we're good to go
+        return {}
```
hoodmane, line 68: Move this guard immediately after `dist = importlib.metadata.Distribution.from_name(...)`

micropip/_commands/check_package_dependencies.py
```diff
@@ -0,0 +1,105 @@
...
+    # Get current list of pyodide requirements
+    requires = dist.read_text("PYODIDE_REQUIRES")
+
+    if requires:
+        depends = json.loads(requires)
+    else:
+        depends = []
+
+    missing = []
+    all_missing = {}
```
hoodmane, line 63: What's the point of `missing` and `all_missing`? This is only called in `freeze` and the return value isn't used, so can't we get rid of these?

micropip/_commands/check_package_dependencies.py
```diff
@@ -0,0 +1,105 @@
...
+    extras (list):
+        List of extras for this package.
+
+    fix_deps (boolean):
+        If this is True, any missing dependencies that are in
+        the current package list will be added to the current
+        package.
+
+    recursive (boolean):
+        If this is True, dependencies of dependencies will be checked.
```
hoodmane, line 35: I'm not that big of a fan of these boolean arguments, especially since we don't use them. If possible it's better to make multiple different functions and compose them to get different behaviors rather than taking arguments.

hoodmane (commented):
Thanks @joemarshall. I made some nitpicky comments.

joemarshall:
Oh and fixed / added tests for freeze.

micropip/_utils.py
```diff
@@ -181,3 +185,75 @@ def platform_to_version(platform: str) -> str:
         )
 
     raise ValueError(f"Wheel interpreter version '{tag.interpreter}' is not supported.")
+
+
+def fix_package_dependencies(package_name, *, extras=None):
```
ryanking13, line 190: Could you please add type hints?

micropip/_utils.py
```diff
@@ -181,3 +185,75 @@ def platform_to_version(platform: str) -> str:
...
+
+    if requires:
+        depends = json.loads(requires)
+    else:
+        depends = []
+
+    if extras is None:
+        extras = [None]
+    else:
+        extras.append(None)
```
ryanking13, line 237: This will modify the original parameter, so maybe instead:

```
extras = extras + [None]
```

ryanking13 (approved):
Thanks @joemarshall!

I have some minor comments otherwise LGTM.

Could you please update the [changelog](https://github.com/pyodide/micropip/blob/main/CHANGELOG.md)? I think we can use a version 0.4.1 and release with this PR.

joemarshall:
Okay, I think I've done those changes okay now

ryanking13:
Thanks @joemarshall! Pyodide 0.24.0 has already been released, so I think this would be included in 0.24.1

hoodmane:
Oh wait this is a micropip change, not sure if these labels make sense.

PR #83, opened 2023-09-11
Change package query order when custom index URLs are given

### LINKED ISSUE #80: My attempt to use index_urls for installing a scikit-learn wheel from pypi.anaconda.org
### Why I want to do this

It would be nice to be able to build a stable and dev wheel for scikit-learn and make it easily installable inside Pyodide. This would allow to control the version of scikit-learn that the user uses, independently of the scikit-learn version provided by the Pyodide distribution.

pypi.anaconda.org seems like a good place since scikit-learn (and other scientific packages) are already using it for some of scikit-learn, for example for hosting nightly scikit-learn wheels see https://anaconda.org/scientific-python-nightly-wheels.

This was mentioned in other places e.g. https://github.com/pyodide/pyodide/issues/3049#issuecomment-1558647194.

### What I tried

I uploaded a scikit-learn wheel to https://pypi.anaconda.org/lesteve/simple/scikit-learn/

I was hoping to be able to use it with `index_urls` (at the same time I expected CORS issues to come into play):

```py
import micropip
micropip.install('scikit-learn', index_urls='https://pypi.anaconda.org/lesteve/simple')
```

Looking at my browser console, I was surprised to realise that the scikit-learn wheel is coming from 
`https://cdn.jsdelivr.net/pyodide/dev/full/` rather than the `index_urls` I specified. In the use case I outlined at the beginning, I would like that `index_urls` takes precedence over what Pyodide provides.

Maybe this is expected since index_urls is only supposed to work with Python wheels and not compiled wheels? Or maybe packages provided by Pyodide have a special status for `micropip`?

Full disclosure: I then tried to build locally a Pyodide dist without scikit-learn and indeed in this case `index_urls` is used but (as kind of expected) there is a CORS issue.

ryanking13:
Resolve #80 

This PR changes the query order of micropip.

1. If `index_urls` is not given
 
Look for `pyodide-lock.json` first, then query PyPI.

2. If `index_urls` is given

Query index URLs first, then look for `pyodide-lock.json`.


CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.4.1] - 2023/09/XX
```
rth, line 9: Let's make it a 0.5.0 as its not really a bug fix, but rather a behavior change.

rth (approved):
Thanks @ryanking13! This is indeed very useful.

PR #85, opened 2023-09-19
CI Run tests with pyodide 0.24.0

ryanking13:
- Update Pyodide version in CI to 0.24.0
- ~Use `response.headers` which is added in Pyodide 0.24.~


.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        pyodide-version: [0.22.1, 0.23.2]
+        pyodide-version: [0.24.0]
```
rth, line 21: Should we document a compatibility matrix somewhere? Of compatible Pyodide version for a given micropip release?

micropip/_compat_in_pyodide.py
```diff
@@ -37,10 +36,7 @@ async def fetch_string_and_headers(
     response = await pyfetch(url, **kwargs)
 
     content = await response.string()
-    # TODO: replace with response.headers when pyodide>= 0.24 is released
-    headers: dict[str, str] = Object.fromEntries(
-        response.js_response.headers.entries()
-    ).to_py()
```
rth, line 43: So it's a bit of a question how fast we drop support for older pyodide releases in micropip.

I'm not opposed to doing this but just something I wonder about, especially that here there is no major benefits from updating (aside from having slightly simpler code in this case).

ryanking13:
@rth Hmm, yes that's a good point.

As you mentioned, there isn't much benefit from dropping Pyodide < 0.24.0 here in this PR, so I'll revert that change, and add some compatibility table.

COMPATIBILITY.md
```diff
@@ -0,0 +1,8 @@
+# Compatible Pyodide versions
+
+Following versions of micropip and Pyodide are tested in CI.
+Other versions may work, however with no guarantee.
```
hoodmane, line 4: Minor language suggestion:
```suggestion
The following versions of micropip and Pyodide are tested in CI.
Other versions may work, however we make no guarantees.
```

ryanking13: Thanks!

PR #86, opened 2023-09-19
MAINT Move `WheelInfo` into a separate file

ryanking13:
This is a maintenance PR that includes the following:

- Move `WheelInfo` class from `transaction.py` to `wheelinfo.py`.
- Explicitly mark private methods of `WheelInfo` as private.
- Improve test coverage of `WheelInfo` class.
- Change `ProjectInfo` that returns from `package_index.query_package` to hold `WheelInfo` directly.


rth (approved):
Thanks for doing this refactoring! Code wise LGTM. 

Looks like there is one missing rename which makes a test fail.

ryanking13:
It looks like there was a regression in recent setuptools_scm release: https://github.com/pypa/setuptools_scm/issues/905

PR #87, opened 2023-09-22
Remove vendored pip and add Metadata class

ryanking13:
This PR is a preparation to support [PEP 658: Serve Distribution Metadata in the Simple Repository API](https://peps.python.org/pep-0658/).

In order to support PEP 658, we need a class that can parse the Distribution Metadata file. For now, this it is implemented [inside the vendored pip](https://github.com/pyodide/micropip/blob/0f0eb116329e7a70d0b50884934ff1afc2e063bb/micropip/externals/pip/_vendor/pkg_resources.py#L388-L427).

The vendored pip code contains a bunch of codes that we are not using anymore, and it does not expose an API for parsing a single metadata file. So this PR removes the vendored pip, and instead adds a class `Metadata` that handles parsing the metadata file and calculating requirements.


micropip/_compat_not_in_pyodide.py
```diff
@@ -64,45 +63,6 @@ def canonicalize_name(name: str) -> str:
     return _canonicalize_regex.sub("-", name).lower()
 
 
-# Vendored from pip
-class UnsupportedWheel(Exception):
-    """Unsupported wheel."""
-
-
-def wheel_dist_info_dir(source: ZipFile, name: str) -> str:
```
ryanking13, line 72: I think there is not much reason to import this from pyodide. I move it to `metadata.py`.

rth:
> re, and it does not expose an API for parsing a single metadata file

Thanks that sounds reasonable. I'm just wondering whether we could more explicitly separate what comes from micropip and what is a class that you added. So that say it would be easier to diff against upstream changes in the future if we decide to sync with changes to those functions in pip. Unless it's a loose adaptation and essentially an independent code base.

ryanking13:
> Unless it's a loose adaptation and essentially an independent code base.

I think this is kind of the way this PR is implemented. I did some copy-paste, but also did some rewrite as well. So if you are okay I would like to merge this as is. Maybe I'll open a follow-up PR that moves methods that come from pip without any modification to the separate file so we can easily distinguish them.

micropip/wheelinfo.py
```diff
@@ -42,16 +40,15 @@ class WheelInfo:
     # Fields below are only available after downloading the wheel, i.e. after calling `download()`.
 
     _data: IO[bytes] | None = None  # Wheel file contents.
-    _dist: Distribution | None = None  # pkg_resources.Distribution object.
+    _metadata: Metadata | None = None  # Wheel metadata.
     _requires: list[Requirement] | None = None  # List of requirements.
 
-    # Note: `_project_name`` is taken from the wheel metadata, while `name` is taken from the wheel filename or metadata of the package index.
-    #       They are mostly the same, but can be different in some weird cases (e.g. a user manually renaming the wheel file), so just to be safe we store both.
-    _project_name: str | None = None  # Project name.
```
rth, line 50: So we would no longer get the name from the wheel metadata. then? I feel it's a safer source than the filename.

rth (approved):
Thanks! A couple of comments otherwise LGTM.

micropip/wheelinfo.py
ryanking13: > no longer get the name from the wheel metadata. then?

Interestingly, we didn't calculate the name from the metadata even before.

The `project_name` was calculated by just escaping the name that was passed by the caller.

https://github.com/pyodide/micropip/blob/1807ccb1a9241ee0023d59d5079a25c89378ad3f/micropip/externals/pip/_vendor/pkg_resources.py#L187

I moved this part to `__post_init__` method. So the behavior is not changed in this PR.

tests/test_data/metadata/urllib3-2.0.5-py3-none-any.whl.metadata
```diff
@@ -0,0 +1,158 @@
+Metadata-Version: 2.1
```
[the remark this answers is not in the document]
ryanking13: Sure I will.

micropip/wheelinfo.py
ryanking13: I'll open a follow-up PR to parse the package name from the metadata file.

PR #90, opened 2023-10-08
Draft: Add PEP-658 support

ryanking13:
This PR adds [PEP 658: Serve Distribution Metadata in the Simple Repository API](https://peps.python.org/pep-0658/) support to micropip, which means that when using Simple API to install a package, micropip will check for distribution metadata and downloads it simultaneously with the wheel to speed up dependency resolution.

I think this PR is mostly ready, but I'll split it into 2-3 PRs, and add more tests (it is kinda tricky to write a test for this), so I'm opening it as a draft.


rth:
Thanks for working on this!

 Related question @ryanking13 : do we do (or plan to do) anything with the `.whl.metadatafiles` we create and distribute via jsdelvir? As currently they are not being served by the CDN https://github.com/jsdelivr/jsdelivr/issues/18531

ryanking13:
> Related question @ryanking13 : do we do (or plan to do) anything with the .whl.metadatafiles we create and distribute via jsdelvir? As currently they are not being served by the CDN https://github.com/jsdelivr/jsdelivr/issues/18531

Yes, that's a good point. I think we'll need to serve .metadata files in the future. If I understand correctly, there has been some work in the direction of allowing the pyodide distribution to be used as a package index. (I don't know what this will eventually look like).

- https://github.com/pyodide/pyodide/pull/3981
- https://github.com/pyodide/pyodide/issues/4182
- https://github.com/pyodide/pyodide/issues/3979

ryanking13:
Superceded by #139

PR #91, opened 2023-10-11
MAINT Store data in bytes not io.BytesIO

ryanking13:
This is a split off of #90.

This PR changes how wheel data is stored in `WheelInfo` class during the package installation: instead of converting the downloaded wheel data to `io.BytesIO` and storing it, store it as `bytes` and convert it to `io.BytesIO` when we need to pass it to other methods that accepts `io.BytesIO` object.

No functional change is intended, just to avoid the mistake of having to do a seek(0) after reading the data.

There seems a slight overhead of converting bytes to io.BytesIO, but I think it is not critical.

<details>
  <summary>Benchmark</summary>

```python
import io
import timeit

num_iterations = 10000
bytes_size = 100_000
bytes_obj = b'0' * bytes_size

def bench_bytes():
    for i in range(num_iterations):
        io_obj = io.BytesIO(bytes_obj)
        io_obj.read()

def bench_bytesio():
    io_obj = io.BytesIO(bytes_obj)
    for i in range(num_iterations):
        io_obj.seek(0)
        io_obj.read()

print('Benchmarking using a single BytesIO object and seek()...')
print(timeit.timeit(bench_bytesio, number=1))
print('Benchmarking converting bytes to io.BytesIO multiple times...')
print(timeit.timeit(bench_bytes, number=1))
```

```
Benchmarking using a single BytesIO object and seek()...
0.0006137999998827581
Benchmarking converting bytes to io.BytesIO multiple times...
0.001084399999854213
```

</details>

### DIFF
--- a/micropip/_compat_in_pyodide.py
+++ b/micropip/_compat_in_pyodide.py
@@ -25,7 +25,7 @@ async def fetch_bytes(url: str, kwargs: dict[str, str]) -> bytes:
         return Path(parsed_url.path).read_bytes()
     if parsed_url.scheme == "file":
         return (await loadBinaryFile(parsed_url.path)).to_bytes()
-    
+
     return await (await pyfetch(url, **kwargs)).bytes()
 
 
--- a/micropip/_compat_in_pyodide.py
+++ b/micropip/_compat_in_pyodide.py
@@ -1,5 +1,4 @@
-from io import BytesIO
-from typing import IO
+from pathlib import Path
 from urllib.parse import urlparse
 
 from pyodide._package_loader import get_dynlibs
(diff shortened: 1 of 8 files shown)


hoodmane (approved):
Thanks @ryanking13. This is a nice improvement. I'd always much rather be handling a `bytes` object than a `BytesIO`

As long as the argument is a `bytes` object, creating a `BytesIO` is effectively free, it just stores the bytes object:
https://github.com/python/cpython/blob/v3.11.6/Modules/_io/bytesio.c#L945-L949
So using them only once rather than having to `seek(0)` all the time is probably a better pattern.

rth:
@ryanking13  Thanks! Let us know if you plan to address Hood's suggestions, otherwise +1 to merge even as is.

ryanking13:
Thanks for the review! I had forgotten about this PR.

PR #94, opened 2023-12-28
TST Add a fixture for testing wheels from external URL

ryanking13:
Adds a pytest fixture that simplifies installing a wheel from external URL.

This replaces a common pattern we use when testing a wheel, which is:

```python
with spawn_web_server(TEST_WHEEL_DIR) as server:
        server_hostname, server_port, _ = server
        url = f"http://{server_hostname}:{server_port}/"
        wheel_url = url+"....whl"

        micropip.install(wheel_url)
```

with:

```python
pkg = test_wheel_catalog.get("<pkg_name>")
pkg.url
micropip.install(pkg_url)
```


tests/conftest.py
```diff
@@ -93,6 +87,68 @@ def selenium_standalone_micropip(selenium_standalone, wheel_path):
...
+        self._wheels[name] = self.Wheel(
+            path, name, path.name, name.replace("-", "_"), url
+        )
+
+    def get(self, name: str) -> Wheel:
+        return self._wheels[name]
+
+
+@pytest.fixture(scope="session")
+def test_wheel_catalog():
```
hoodmane, line 143: Maybe just call it `wheel_catalog`?

tests/test_install.py
```diff
@@ -29,23 +29,20 @@ def test_install_simple(selenium_standalone_micropip):
...
-        )
+    selenium.run_js(
+        f"""
+        await pyodide.runPythonAsync(`
+            import micropip
+            await micropip.install('{url}')
+            import snowballstemmer
+        `);
+        """
+    )
```
hoodmane, line 45: Maybe while we're att it we should transform this one to use `@run_in_pyodide`:

```suggestion
    @run_in_pyodide
    def install_from_url(url):
        import micropip
        await micropip.install(url)
        import snowballstemmer

    install_from_url(url)
```

hoodmane (approved):
Thanks for doing this maintenance work @ryanking13!

ryanking13:
Thanks for the review!

PR #97, opened 2024-01-29
Fix micropip.install to search shared libraries inside the wheel correctly

ryanking13:
Previously, micropip was not able to locate shared libraries in the `.libs` directory of the wheel.

For instance, in Pyodide 0.25.0,

```js
await pyodide.loadPackage("Shapely")
```

works, while

```sh
>>> import micropip
>>> await micropip.install("https://cdn.jsdelivr.net/pyodide/v0.25.0/full/Shapely-1.8.
2-cp311-cp311-emscripten_3_1_46_wasm32.whl")
  File "/lib/python3.11/site-packages/micropip/_commands/install.py", line 176, in ins
tall
    await asyncio.gather(*wheel_promises)
  File "/lib/python3.11/site-packages/micropip/transaction.py", line 168, in install
    await self.load_libraries(target)
  File "/lib/python3.11/site-packages/micropip/transaction.py", line 159, in load_libr
aries
    await asyncio.gather(*map(lambda dynlib: loadDynlib(dynlib, False), dynlibs))
pyodide.ffi.JsException: Error: Didn't expect to load any more file_packager files!
```

does not work.

This is because `micropip.install` does not search the path that `pyodide.loadPackage` looks for.

This PR fixes it by using the same API `pyodide.loadPackage` is using when loading shared libraries.

- [ ] changelog


tests/conftest.py
```diff
@@ -127,10 +127,13 @@ def _register_handler(self, path: Path) -> str:
 
         return self._httpserver.url_for(f"/{path.name}")
 
-    def add_wheel(self, path: Path):
+    def add_wheel(self, path: Path, replace: bool = True):
         name = parse_wheel_filename(path.name)[0]
         url = self._register_handler(path)
 
+        if name in self._wheels and not replace:
+            return
```
hoodmane, line 135: Maybe we should raise here? I think silently doing nothing is likely to be confusing.

.github/workflows/main.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        pyodide-version: [0.22.1, 0.23.4, 0.24.0]
+        pyodide-version: [0.24.1, 0.25.0]
```
hoodmane, line 21: Does this change set rely on 0.24? Looks to me like `loadDynlibsFromPackage` was added in 0.22 and hasn't been changed since. Also, how many Pyodides does micropip target to support? Rolling support for last two versions? and same for Pyodide supporting old micropip versions?

hoodmane (approved):
Thanks @ryanking13! Could you add a changelog entry?

.github/workflows/main.yml
ryanking13: It relies on the list methods of PyProxy, which was added in 0.24 (https://github.com/pyodide/pyodide/commit/1aea59f55a1bfa0a790bebc99480e006553fe5df).

> Also, how many Pyodides does micropip target to support? Rolling support for last two versions? and same for Pyodide supporting old micropip versions?

My opinion is to ensure backwards compatibility whenever possible, but make breaking changes when necessary. Maybe we can make a compatibility table for micropip, just as how we do in pytest-pyodide.

hoodmane: Is there a reason that you're dropping 0.22 and 0.23?

hoodmane: Oops should have read more carefully.

tests/conftest.py
ryanking13: I used this method to register all the wheels included in the Pyodide distribution as test wheels, as without this, they would conflict with pytest wheel, etc that I put in for testing.

I think it will be fine as long as we know what we are doing. I added some comments about it.

ryanking13:
Oh, I forgot to update the changelog...

PR #109, opened 2024-05-24
Index pip-installed packages

### LINKED ISSUE #107: Micropip.freeze no work?
```
pip install pyodide-build
pyodide venv .venv-pyodide
.venv-pyodide/bin/pip install fastapi
.venv-pyodide/bin/python -c 'import micropip; print(micropip.freeze(), file=open("pyodide-lock.json","w"))'
.venv-pyodide/bin/pip freeze | grep fastapi # finds fastapi
cat pyodide-lock.json | grep fastapi # no fastapi??
```

hoodmane:
This isn't very efficient, we should use pep 658 metadata which pypi provides now so we don't have to redownload the wheel. Also we should do fetches concurrently and asyncio.gather them. But it gets the basic idea.

Would resolve #107.

cc @jaraco @ryanking13

### DIFF
--- a/micropip/_commands/freeze.py
+++ b/micropip/_commands/freeze.py
@@ -1,12 +1,110 @@
 import importlib.metadata
 import json
+import sys
 from copy import deepcopy
-from typing import Any
+from importlib.metadata import Distribution
+from typing import TypedDict
 
 from packaging.utils import canonicalize_name
+from packaging.version import Version
 
 from .._compat import REPODATA_INFO, REPODATA_PACKAGES
 from .._utils import fix_package_dependencies
+from ..package_index import query_package
+
+IN_VENV = sys.prefix != sys.base_prefix
+
+
+class PkgEntry(TypedDict):
+    name: str
+    version: str
+    file_name: str
+    install_dir: str
+    sha256: str | None
+    imports: list[str]
+    depends: list[str]
+
+
+def get_pkg_entry_micropip(dist: Distribution, url: str) -> PkgEntry:
+    name = dist.name
+    version = dist.version
+    sha256 = dist.read_text("PYODIDE_SHA256")
+    assert sha256
+    imports = (dist.read_text("top_level.txt") or "").split()
+    requires = dist.read_text("PYODIDE_REQUIRES")
+    if not requires:
+        fix_package_dependencies(name)
+        requires = dist.read_text("PYODIDE_REQUIRES")
+    if requires:
+        depends = json.loads(requires)
+    else:
+        depends = []
+
+    return dict(
+        name=name,
+        version=version,
+        file_name=url,
+        install_dir="site",
+        sha256=sha256,
+        imports=imports,
+        depends=depends,
+    )
+
+
+async def get_pkg_entry_pip(dist: Distribution) -> PkgEntry | None:
+    resp = await query_package(dist.name)
+    ver = resp.releases.get(Version(dist.version), None)
+    if ver is None:
+        return None
+    wheel = next(ver)
+    await wheel.download({})
+    requires = [req.name for req in wheel.requires(set())]
+    return dict(
+        name=dist.name,
+        version=dist.version,
+        file_name=wheel.url,
+        install_dir="site",
+        sha256=wheel.sha256,
+        imports=[],
+        depends=requires,
+    )
+
+
+async def freeze2() -> str:
+    """Produce a json string which can be used as the contents of the
+    ``repodata.json`` lock file.
+
+    If you later load Pyodide with this lock file, you can use
+    :js:func:`pyodide.loadPackage` to load packages that were loaded with :py:mod:`micropip`
+    this time. Loading packages with :js:func:`~pyodide.loadPackage` is much faster
+    and you will always get consistent versions of all your dependencies.
+
+    You can use your custom lock file by passing an appropriate url to the
+    ``lockFileURL`` of :js:func:`~globalThis.loadPyodide`.
+    """
+    packages = deepcopy(REPODATA_PACKAGES)
+    for dist in importlib.metadata.distributions():
+        name = dist.name
+        url = dist.read_text("PYODIDE_URL")
+        if url:
+            pkg_entry = get_pkg_entry_micropip(dist, url)
+        elif IN_VENV:
+            res = await get_pkg_entry_pip(dist)
+            if not res:
+                continue
+            pkg_entry = res
+        else:
+            continue
+
+        packages[canonicalize_name(name)] = pkg_entry
+
+    # Sort
+    packages = dict(sorted(packages.items()))
+    package_data = {
+        "info": REPODATA_INFO,
+        "packages": packages,
+    }
+    return json.dumps(package_data)
 
 
 def freeze() -> str:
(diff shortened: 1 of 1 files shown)


micropip/_commands/freeze.py
```diff
@@ -1,12 +1,110 @@
...
+        depends=depends,
+    )
+
+
+async def get_pkg_entry_pip(dist: Distribution) -> PkgEntry | None:
+    resp = await query_package(dist.name)
+    ver = resp.releases.get(Version(dist.version), None)
+    if ver is None:
+        return None
+    wheel = next(ver)
```
jaraco, line 59: This approach seems risky. What if the package doesn't have a wheel, but only an sdist? What if the sdist is chosen over the wheel? What if the package was installed from a local checkout that advertises the same version (but is different)? What if the package was installed from a local checkout that advertises a local version (that's not present in PyPI)?

micropip/_commands/freeze.py
```diff
@@ -1,12 +1,110 @@
...
+    You can use your custom lock file by passing an appropriate url to the
+    ``lockFileURL`` of :js:func:`~globalThis.loadPyodide`.
+    """
+    packages = deepcopy(REPODATA_PACKAGES)
+    for dist in importlib.metadata.distributions():
+        name = dist.name
+        url = dist.read_text("PYODIDE_URL")
+        if url:
+            pkg_entry = get_pkg_entry_micropip(dist, url)
+        elif IN_VENV:
```
jaraco, line 91: I'm unsure why IN_VENV is relevant. Packages can be installed without virtualenvs (e.g. on PYTHONPATH). And virtualenvs can expose site-packages.

What is the motivation for varying on IN_VENV?

micropip/_commands/freeze.py
```diff
@@ -1,12 +1,110 @@
...
+    )
+
+
+async def get_pkg_entry_pip(dist: Distribution) -> PkgEntry | None:
+    resp = await query_package(dist.name)
+    ver = resp.releases.get(Version(dist.version), None)
+    if ver is None:
+        return None
+    wheel = next(ver)
+    await wheel.download({})
```
jaraco, line 60: I see this method doesn't yet exist. This call could be very expensive and would need to manage the lifecycle of the downloaded content. It may also need to re-implement things that pip does like caching or re-using pip's cache.

jaraco (commented):
This is an interesting idea, but as you point out, the querying and download of wheels seems expensive. Have we yet explored how important it is for the `file_name` and `sha256` fields to be populated in the result? Could those fields instead be made optional for installed packages that no longer have their attribution? i.e. Why not make `file_name: str | None`?

jaraco:
See #110 where I've done some work to refactor the functionality to make it easier to alter behavior such as proposed here.

micropip/_commands/freeze.py
```diff
@@ -1,12 +1,110 @@
...
+    requires = [req.name for req in wheel.requires(set())]
+    return dict(
+        name=dist.name,
+        version=dist.version,
+        file_name=wheel.url,
+        install_dir="site",
+        sha256=wheel.sha256,
+        imports=[],
```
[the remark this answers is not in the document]
hoodmane: It's a bit unfortunate that `top_level.txt` is setuptools only. As I understand it, the other build backends don't have anything similar. Do you know why?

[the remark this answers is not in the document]
ryanking13: I think `top_level.txt` is not very good for querying import names, and probably we should stop using it in micropip.

Ref: https://github.com/pyodide/pyodide/pull/3006#pullrequestreview-1081606287

micropip/_commands/freeze.py
```diff
@@ -1,12 +1,110 @@
...
+        return None
+    wheel = next(ver)
+    await wheel.download({})
+    requires = [req.name for req in wheel.requires(set())]
+    return dict(
+        name=dist.name,
+        version=dist.version,
+        file_name=wheel.url,
+        install_dir="site",
+        sha256=wheel.sha256,
```
ryanking13, line 67: So IIUC, we are fetching the installed package again to get the URL and the sha256? I am not sure if is it a reliable behavior. What happens if a user installed packages from non-PyPI repositories or from URL?

hoodmane: It works if they installed from url (at least in principle it can be made to work) because we can get the url from `direct_url.json`. The multiple indexes problem is bigger because we don't have metadata indicated which index we got it from. I think micropip should store the index url into maybe `INDEX`.

@pradyunsg @henryiii is there something I'm missing? If someone has a pip.conf with an extra-index-url or passes this as a cli instruction, is it possible to reconstruct which packages in the venv came from which indices? If not maybe we could make a packaging PEP adding INDEX to package metadata.

micropip/_commands/freeze.py
jaraco: Pip does expose a `list` command that will list all matching packages in the cache:

```
 ~ @ pip cache list foo.bar-0.2 --format abspath
/Users/jaraco/Library/Caches/pip/wheels/0b/ec/8b/f669ae78d94a691e6c63872909196857158b3c20f045058e7b/foo.bar-0.2-py3-none-any.whl
/Users/jaraco/Library/Caches/pip/wheels/10/f7/18/a4a94460da8bf74f841d7df4f10984b730907bea49783615a4/foo.bar-0.2-py3-none-any.whl
/Users/jaraco/Library/Caches/pip/wheels/11/9f/15/3a750be9586c80770d8355ac4e9dc3c58237b1d994f36a2a21/foo.bar-0.2-py3-none-any.whl
/Users/jaraco/Library/Caches/pip/wheels/1b/9d/16/254558b9de54233baf2ef9a35eac16ad3c23a8c5fe4122374f/foo.bar-0.2-py3-none-any.whl
/Users/jaraco/Library/Caches/pip/wheels/20/96/a7/21d787cb6f024ae978823ce36306736a68ffc7c728077edf4d/foo.bar-0.2-py3-none-any.whl
...
```

It does appear as if that "wheel cache" is only for locally-built wheels.

I haven't found a way to query for pip's http cache. Here's what pip gives for `cache info`:

```
 draft @ pip cache info
Package index page cache location (pip v23.3+): /Users/jaraco/Library/Caches/pip/http-v2
Package index page cache location (older pips): /Users/jaraco/Library/Caches/pip/http
Package index page cache size: 883.2 MB
Number of HTTP files: 10736
Locally built wheels location: /Users/jaraco/Library/Caches/pip/wheels
Locally built wheels size: 21.2 MB
Number of locally built wheels: 440
```

It's possible the `http-v2` cache is manifest using some library. If not, there's seemingly no API to access it. My guess, based on the fact that there are two cache locations, is that the http-v2 cache is reasonably stable.

micropip/_commands/freeze.py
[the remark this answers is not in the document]
jaraco: Several good examples in that thread. Also, namespace packages aren't reflected accurately either:

```
 @ pip-run jaraco.functools -- -c "import importlib.metadata as md; print(md.distribution('jaraco.functools').read_text('top_level.txt'))"
jaraco
```

(jaraco.functools is one of many packages that exposes the top-level `jaraco` package; the unique offering of `jaraco.functools` is the python package of the same name)

PR #111, opened 2024-05-26
Abort other fetches when resolution fails

CNSeniorious000:
For now, if you installs a incompatible package like `jupyter` without `keep_going=True`, it raises after about 1-2s, but if you open the DevTools, you will find that fetches and resolution continues to run although it is already useless.

This is a waste of MBs and CPU. So I propose cancelling other promises after raising.

Reproduction:

```py
from micropip import install
await install("jupyter")
```


CNSeniorious000:
Wow the spelling check is great

CNSeniorious000:
By the way

https://github.com/CNSeniorious000/micropip/blob/703fc54f71c0d666a10e7aeb0353f1364f56aca7/micropip/_compat_in_pyodide.py#L31-L39

micropip/_compat_in_pyodide.py
```diff
@@ -21,6 +22,27 @@
...
+        kwargs["signal"] = controller.signal
+
+        async def fetch_with_abort():
+            try:
+                return await pyfetch(urls, **kwargs)
+            except CancelledError:
+                controller.abort()
+                raise
+
+        return await fetch_with_abort()
```
CNSeniorious000, line 41: By the way, I think maybe this logic can be written into `pyfetch` itself, because `cancel` is more pythonic than using a `AbortController`, and as I understand it, `pyfetch` is a pythonic wrapper for `js.fetch`

hoodmane: PR welcome!

micropip/_compat_in_pyodide.py
```diff
@@ -21,6 +22,27 @@
...
+
+        controller = AbortController.new()
+        kwargs["signal"] = controller.signal
+
+        async def fetch_with_abort():
+            try:
+                return await pyfetch(urls, **kwargs)
+            except CancelledError:
+                controller.abort()
+                raise
```
hoodmane, line 39: Move this into a function taking controller, urls, kwargs as arguments instead of a closure?

micropip/transaction.py
```diff
@@ -52,7 +52,17 @@ async def gather_requirements(
...
+        futures: list[asyncio.Future] = []
+        try:
+            for coro in requirement_promises:
+                futures.append(asyncio.ensure_future(coro))
+            await asyncio.gather(*futures)
+        except ValueError:
+            if not self.keep_going:
+                for future in futures:
+                    if not future.done():
+                        future.cancel()
```
hoodmane, line 64: I haven't investigated how cancellation interacts with our event loop and I'm slightly worried that it could not interact well. But if this passes the tests then we should merge.

hoodmane (approved):
Thanks @CNSeniorious000!

micropip/transaction.py
CNSeniorious000: Wait a second, I just find that this part seems to raises several `asyncio.exceptions.InvalidStateError: invalid state` exceptions

hoodmane: @Zac-HD any tips on how to make my event loop support cancellation? Where to look who to ask etc?

hoodmane: Anyways I'll open a separate issue about this.

micropip/_compat_in_pyodide.py
CNSeniorious000: This is done. and `urls` is a typo so I [amend the commit](https://github.com/pyodide/micropip/compare/703fc54f71c0d666a10e7aeb0353f1364f56aca7..e01491f5051e8b03752563f18ddd425644e6ecc3)

micropip/transaction.py
hoodmane: https://github.com/pyodide/pyodide/issues/4800

ryanking13 (commented):
Can we add some tests? Or would it be too difficult to simulate network failures?

CNSeniorious000:
It's not about network failure, it's about installing pyodide-incompatible packages.

For example, `jupyter` is a huge package and is not installable in pyodide. If you try to `micropip.install("jupyter")`, it will fail soon due to "ValueError: Can't find a pure Python 3 wheel".

But the fetches and resolution still runs in the background, which is wasting network and CPU usage.

So I came into an idea that I may `cancel` all the running futures, to stop resolution after any one resolution failed. (only when keep_going=False because if you want to keep_going, we must resolve all the dependencies, so we can't early-stop)

This is implemented in this PR I think, you can test it by host the wheel using a static server and use the load `http://localhost:.../micropip-...-...whl` in the pyodide console, and try install `jupyter` again. You will see all the fetches got aborted when one resolution fails, as expected.

But you will see the pyodide console print many `asyncio.exceptions.InvalidStateError: invalid state` lines. I don't know why. Maybe this is an issue of the webloop.

hoodmane:
I'll try and check this out and debug the `InvalidStateError` issues...

micropip/_compat_in_pyodide.py
```diff
@@ -21,6 +22,24 @@
...
+            return await pyfetch(url, **kwargs)
+
+        controller = AbortController.new()
+        kwargs["signal"] = controller.signal
+
+        try:
+            return await pyfetch(url, **kwargs)
+        except CancelledError:
+            controller.abort()
+            raise
```
hoodmane, line 38: Would be great to implement this into pyfetch directly in a followup.

micropip/transaction.py
hoodmane: Could you try using a TaskGroup? Maybe that will fix our problems.
https://docs.python.org/3/library/asyncio-task.html#asyncio.TaskGroup

CNSeniorious000: Done, I've refactored this section using `TaskGroup`, now all the futures should be gathered in one top-level `TaskGroup`.

Note that this makes the traceback looks a bit different because `TaskGroup` would group all the exceptions into a `ExceptionGroup`.

micropip/_compat_in_pyodide.py
CNSeniorious000: So should I remove changes in this file and open another PR on `pyfetch` in pyodide/pyodide?

Do you think this feature should provide a parameter for user to opt-out this behavior? Maybe adding a kw-only parameter like `abort_on_cancel=True`?

micropip/transaction.py
CNSeniorious000: The test fails because the raised error is not a `ValueError` but a `ExceptionGroup` containing a `ValueError`.

It's possible to recover the original ValueError by wrap the async with block with a try-except:

```py
    try:
        async with asyncio.TaskGroup() as tg:
            self.tg = tg  # only one task group from the top level
            for requirement in requirements:
                tg.create_task(self.add_requirement(requirement))
    except ExceptionGroup as e:
        raise e.exceptions[-1] from None
```

micropip/_compat_in_pyodide.py
hoodmane: No, let's add it here and also upstream. 

> Maybe adding a kw-only parameter like `abort_on_cancel=True`?

Nah it feels like we're switching from a less correct behavior to a more correct one with this. Let's not overcomplicate the API by forcing people to opt into what they expect.

hoodmane:
Looks like `test_fetch_wheel_fail` has to be updated to expect an ExceptionGroup to be raised.

CNSeniorious000:
I recorded a video to demo this bug, which demonstrates how to reproduce this behavior. Hope that is more clear than my words.

https://github.com/pyodide/micropip/assets/74518716/da0f1e76-674c-4471-a6b0-05fea1f64c43

hoodmane:
I think the problem is you're trying to cancel a future that has already been resolved. There are only three spots in the Python interpreter where `InvalidStateError("invalid state")` is raised:
https://github.com/python/cpython/blob/main/Modules/_asynciomodule.c?plain=1#L557
https://github.com/python/cpython/blob/main/Modules/_asynciomodule.c?plain=1#L577

hoodmane:
I still think it could be due to a bug in WebLoop.

hoodmane:
One way to learn more about it would be to path the interpreter to log some extra info when it sets the `InvalidStateError`. I'll try to look into it tomorrow.

CNSeniorious000:
> Looks like `test_fetch_wheel_fail` has to be updated to expect an ExceptionGroup to be raised.

I have doubts about this. when `keep_going=True` it still raises a `ValueError`. Do you think it is reasonable to raise a `ExceptionGroup` when `keep_going=False` but `ValueError` otherwise?

I personally prefer the original behavior simply using `gather` because we are constructing the error message ourselves, so we need no `ExceptionGroup`.

IIRC, The point of `TaskGroup` is not only canceling other futures when one fails, but also waiting them and group their exceptions into one. I think we don't need this. We only need the first `ValueError` if `keep_going=True`, else we need all the error messages to construct a single ValueError with meaningful message. So I think maybe `TaskGroup` is not that suitable for our use case?

micropip/_compat_in_pyodide.py
CNSeniorious000, line 1: I realized that the original implementation can only abort in `await pyfetch(...)` stage, but we need to abort `await response.bytes()` too, so I changed the approach.

Now I used a decorator to dependency-inject an `AbortSignal` into the decorated function, and pass that to the call to `pyfetch`. After decorating, the signature of them are the same as before. But maybe this needs a re-review.

CNSeniorious000:
Everything works fine now. Here is a short video for demonstration:

https://github.com/pyodide/micropip/assets/74518716/88ee1cf3-3417-4df8-bfb0-c7c3c5d66f09

But I don't know how to implement a testcase to ensure this behavior.

micropip/_compat_in_pyodide.py
CNSeniorious000: Here I still changed `fetch_bytes` and `fetch_string_and_headers`. Maybe you think that isn't elegant.

In fact, through a more hacking way, it is possible to only decorate it, without dependency injection, and no longer need to pass `signal=signal` in pyfetch themselves, which enable this decorater to be used elsewhere too.

In the `_abort_on_cancel` decorator, replace the input function's `__locals__` with a `ChainMap`. In that context, insert a `_signal` into that namespace, and replace the `pyfetch` in that namespace to `partial(pyfetch, signal=_signal)`. Then we can simplify the patching code:

```diff
@_abort_on_cancel
async def fetch_string_and_headers(
-   signal: AbortSignal, url: str, kwargs: dict[str, str]
+   url: str, kwargs: dict[str, str]
) -> tuple[str, dict[str, str]]:
-   response = await pyfetch(url, **kwargs, signal=signal)
+   response = await pyfetch(url, **kwargs)
```

potential downsides:
1. the code may become a little confusing
2. if the user don't use the name `pyfetch`, but using other names, it won't be patched as `partial(pyfetch, signal=_signal`

hoodmane:
Next week I'll try to see if I can make a test. I think we can monkeypatch fetch perhaps...

Issue #112, opened 2024-06-03
Better modularization + dependency injection


ryanking13:
micropip is used in many different downstream projects and has many different requirements, so I think it would be nice to 1) modularize overall micropip behavior, and 2) use dependency injection to customize micropip behavior.

## `micropip.Micropip()`

Currently, micropip only has a global state and can only be used with global APIs like `micropip.install` and `micropip.freeze`.
This issue proposes to define a `Micropip` class in addition to this, and make it possible to create instances with local state.

The existing APIs will still be available (no breaking changes), and they will internally reference the default Micropip instance.

```python
await micropip.install("...")  # this will still work

# Additionally, we have this.
micropip1 = micropip.Micropip(param1, param2, ...)
await micropip1.install()
```

## Dependency injection

When instantiating the Micropip instance, we can accept several parameters including, `resolver`, `installer`, etc.

```python
def Micropip:
    def __init__(self, resolver, installer, etc):
        self.resolver = resolver
        self.installer = installer

    def dependency_resoltion(self, package):
        self.resolver.do_dependency_resolution(package)

    def install_package(self, package):
        self.installer.do_install_package(package)
```

This will make it easier to tweak micropip's behavior, which will make it easier for us to implement new features, or people can implement their own features. For instance, someone can make a backtracking resolver and replace our simple resolver if they want, or it might be even possible to use micropip in non-Pyodide environments.

## Required works

The global state of micropips, which is currently jumbled in many ways, needs to be better categorized by purpose and defined in a clear API.

hoodmane:
cc @jaraco

jaraco:
This approach sounds reasonable. I like the idea of essentially creating hooks that a programmer can override. My guess is that these hooks are going to demand some pretty intimate access to the state in Micropip. To the extent they can be stateless functions, that'd be great, but doing so would drastically limit the ability of that signature to evolve. The first step is obviously going to be to dogfood the concept with the standard implementation. It'll be interesting to see what interfaces emerge from that work and how clean those interfaces can be.

Another possible way for programmers to extend/override functionality could be to offer overridable methods, e.g.:

```python
class Micropip:
    def resolve(self, package):
        ...

    def install(self, package):
        ...

class Backtracking:
    """Mix-in for Micropip to implement a backtracking resolver."""
    def resolve(self, package):
        # doesn't call super()
        ...


class BacktrackingMicropip(Backtracking, Micropip):
    pass
```

[...]

PR #114, opened 2024-06-16
Modularized client for state management

RulerOfCakes:
This PR adds a `Manager` class to act as a state encapsulation object for managing the state of `micropip` in a local `Manager` instance. 

The initial goal of this PR is to have the exposed [commands](https://github.com/pyodide/micropip/tree/main/micropip/_commands) API be usable through the `Manager`, enabling having multiple instances of it with different local state. The current global API will remain supported, but the 'global state' should be independent of other alive `Manager` instances.

### DIFF
--- a/micropip/manager.py
+++ b/micropip/manager.py
@@ -1,0 +1,34 @@
+class Manager:
+    """
+    Manager provides an extensible interface for customizing micropip's behavior.
+
+    Each Manager instance holds its own local state that is 
+    independent of other instances, including the global state.
+    """
+
+    def __init__(self):
+        pass
+
+    def install(self):
+        pass
+
+    def list(self):
+        pass
+
+    def freeze(self) -> str:
+        pass
+
+    def add_mock_package(self):
+        pass
+
+    def list_mock_packages(self):
+        pass
+
+    def remove_mock_package(self):
+        pass
+
+    def uninstall(self):
+        pass
+
+    def set_index_urls(self, urls: list[str] | str) -> None:
+        pass
--- a/micropip/manager.py
+++ b/micropip/manager.py
@@ -23,25 +23,25 @@ class Manager:
         pass
 
     def install(self):
-        pass
+        raise NotImplementedError()
 
     def list(self):
-        pass
+        raise NotImplementedError()
 
-    def freeze(self) -> str:
-        pass
+    def freeze(self):
+        raise NotImplementedError()
 
     def add_mock_package(self):
-        pass
+        raise NotImplementedError()
 
     def list_mock_packages(self):
-        pass
+        raise NotImplementedError()
 
     def remove_mock_package(self):
-        pass
+        raise NotImplementedError()
 
     def uninstall(self):
-        pass
+        raise NotImplementedError()
 
     def set_index_urls(self, urls: List[str] | str):
         """
--- a/micropip/manager.py
+++ b/micropip/manager.py
@@ -1,4 +1,7 @@
-from typing import Any, List
+from typing import (  # noqa: UP035 List import is necessary due to the `list` method
+    Any,
+    List,
+)
 
 from micropip import package_index
 
@@ -43,7 +46,7 @@ class Manager:
     def uninstall(self):
         raise NotImplementedError()
 
-    def set_index_urls(self, urls: List[str] | str):
+    def set_index_urls(self, urls: List[str] | str):  # noqa: UP006
         """
         Set the index URLs to use when looking up packages.
 
--- a/micropip/manager.py
+++ b/micropip/manager.py
@@ -1,12 +1,19 @@
+from typing import List
+from micropip import package_index
+
+
 class Manager:
     """
     Manager provides an extensible interface for customizing micropip's behavior.
 
-    Each Manager instance holds its own local state that is 
+    Each Manager instance holds its own local state that is
     independent of other instances, including the global state.
+
+    TODO: Implement all of the following global commands to utilize local state.
     """
 
     def __init__(self):
+        self.index_urls = package_index.DEFAULT_INDEX_URLS
         pass
 
     def install(self):
@@ -30,5 +37,27 @@ class Manager:
     def uninstall(self):
         pass
 
-    def set_index_urls(self, urls: list[str] | str) -> None:
-        pass
+    def set_index_urls(self, urls: List[str] | str):
+        """
+        Set the index URLs to use when looking up packages.
+
+        - The index URL should support the
+            `JSON API <https://warehouse.pypa.io/api-reference/json/>`__ .
+
+        - The index URL may contain the placeholder {package_name} which will be
+            replaced with the package name when looking up a package. If it does not
+            contain the placeholder, the package name will be appended to the URL.
+
+        - If a list of URLs is provided, micropip will try each URL in order until
+            it finds a package. If no package is found, an error will be raised.
+
+        Parameters
+        ----------
+        urls
+            A list of URLs or a single URL to use as the package index.
+        """
+
+        if isinstance(urls, str):
+            urls = [urls]
+
+        self.index_urls = urls[:]
--- a/micropip/manager.py
+++ b/micropip/manager.py
@@ -1,4 +1,5 @@
 from typing import Any, List
+
 from micropip import package_index
 
 
--- a/micropip/manager.py
+++ b/micropip/manager.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import Any, List
 from micropip import package_index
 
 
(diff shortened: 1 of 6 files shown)


micropip/manager.py
```diff
@@ -0,0 +1,69 @@
...
+    independent of other instances, including the global state.
+
+    TODO: Implement all of the following global commands to utilize local state.
+    """
+
+    def __init__(self):
+        self.index_urls = package_index.DEFAULT_INDEX_URLS
+
+        # TODO: initialize the compatibility layer
+        self.repodata_packages: dict[str, dict[str, Any]] = {}
```
ryanking13, line 20: `repodata` is a legacy name. We used a lock file called `repodata.json,` but we've changed its name to `pyodide-lock.json,` so it would probably be better to rename this variable too.

(no action required in this PR)

ryanking13 (approved):
Thanks for working on this @RulerOfCakes!

micropip/manager.py
```diff
@@ -0,0 +1,69 @@
...
+    def list_mock_packages(self):
+        pass
+
+    def remove_mock_package(self):
+        pass
+
+    def uninstall(self):
+        pass
+
+    def set_index_urls(self, urls: List[str] | str):
```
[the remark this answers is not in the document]
RulerOfCakes: This was because of the name conflict with the `list()` method, unlike the `_list` export from `_commands.list.py` I couldn't alias the method name to something else. My LSP was being a bit weird about using `list` here but I suppose its a local issue..

RulerOfCakes:
I presume the next logical step would be to encapsulate the compatibility layer states(`repodata_packages`, `repodata_info`) within the `Manager` class's initialization and inject them in the dependent remaining commands? I initially had left this PR as draft because I didn't want to make too many changes at once before replacing the global state with a default `Manager` instance.

micropip/manager.py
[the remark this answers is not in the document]
ryanking13: I see, that's funny. It was probably a bad decision to call it `list` at the first time. Anyway, it's not an issue. So let's keep this as-is.

micropip/manager.py
```diff
@@ -0,0 +1,69 @@
...
+        self.index_urls = package_index.DEFAULT_INDEX_URLS
+
+        # TODO: initialize the compatibility layer
+        self.repodata_packages: dict[str, dict[str, Any]] = {}
+        self.repodata_info: dict[str, str] = {}
+
+        pass
+
+    def install(self):
+        pass
```
ryanking13, line 26: Let's put `raise NotImplementedError()` everywhere, to let people now it is under development.

ryanking13 (commented):
> encapsulate the compatibility layer states(repodata_packages, repodata_info) within the Manager class's initialization and inject them in the dependent remaining commands?

Yes, that would be great for the next step. I am okay with merging this PR with this dummy class so you can continue next steps (please check the pre-commit issue though).

I think `micropip.freeze` can be a good candidate for the first command to implement. It only relies on two compatibility objects: `repodata_packages` and `repodata_info`.

micropip/manager.py
```diff
@@ -0,0 +1,69 @@
+from typing import Any, List
+
+from micropip import package_index
+
+
+class Manager:
```
hoodmane, line 6: Do we like the name `Manager`? It's a bit vague but seems alright to me. We should consider whether we can come up with something better. @jaraco maybe you have some thoughts?

micropip/manager.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Manager provides an extensible interface for customizing micropip's behavior.
+
+    Each Manager instance holds its own local state that is
+    independent of other instances, including the global state.
```
hoodmane, line 11: > including the global state.

Well it's not really global then. It's a bit unclear how this works since the result of installing stuff with micropip is generally files placed in the file system. I guess different `Manager` instances can have different install directories?

micropip/manager.py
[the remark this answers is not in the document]
hoodmane: Add a comment next to the `from typing import List` please!

hoodmane (approved):
Thanks @rulerofcakes seems like a reasonable starting point.

micropip/manager.py
RulerOfCakes: I meant to phrase it in a way that the global manager instance(representing the 'global state' for the exposed commands) is also included in the statement, but maybe it isn't necessary to specify it as such.

I would guess that the 'right way' to keep things independent would be to also have different install directories, but I'm not sure how nicely that would play out with sitepackages, or perhaps a mechanism to prevent managers having duplicate install directories. Could you provide your opinion on this? cc: @ryanking13

micropip/manager.py
jaraco: I don't love "Manager", but it's adequate.

I notice below the `repodata_` prefix on a couple of member variables, suggesting that maybe this object is representing `RepoData` or `State` or `RepositoryState` or `Repository` or `Repo` (and maybe the `repodata_` prefix can be removed). I also see that "repodata" is a legacy artifact and "lock" is relevant, so maybe `LockState` or some variant.

Since I see actions about `.install`, and `.freeze`, I'm thinking this is relevant to an `Environment` or `Site`.

I'm liking `Repository` or `Environment` best, but I'd be okay with any option. Feel free to pick/promote one you like best.

Am I right in thinking this class is an internal implementation detail (not part of a public API)? If so, the naming isn't critical and can be changed later.

RulerOfCakes: My thoughts were that eventually this class will become representative of the current public API, with additional extensibility such as dependency injection as discussed in #112, although I would guess having a separate exposed layer and keeping the `Manager` internal is also a feasible choice. (Assuming that the most common case of a user wanting to instantiate their own `Manager` without accessing the current public APIs would be to utilize such extensibility)

I was thinking that `Client` or `Micropip` could suit it better but `Manager` seemed closest to what it was going to do: state management. I do agree its a vague name, and `Repository` also sounds alright too.

ryanking13: Yep, I plan to make this class public. People would do `mgr = micropip.Manager(); mgr.install(...)`.

I prefer `Manager` or `Environment`. The former emphasizes that this is a package manager of sorts, while the latter recognizes that each instance is like a virtual environment of sorts.

Let's go with `Manager` for now. We still have time before implementing all these APIs and making this class public.

micropip/manager.py
ryanking13: > I guess different Manager instances can have different install directories?

Ideally, we may try to have each instance behave completely independently as if each instance becomes a virtual environment. For example, by combining micropip and subinterpreter, we may be able to support different interpreters to install and run different sets of packages.

However, I think most of the time, people will end up using one instance, and we would recommend doing so, too.

ryanking13:
Could you please check the linter issues? I think we can just add `#noqa : ...` for now.

```
micropip/_commands/install.py:133: error: Argument "ctx" to "Transaction" has incompatible type "Environment"; expected "dict[str, str]"  [arg-type]
```

This one is not related to your PR but to `packaging` update I guess. I think we can ignore it for now too.

micropip/manager.py
hoodmane: Maybe call it `PackageManager` then? or `???Environment`? I feel like we could use an extra word in the name.

micropip/manager.py
hoodmane: This is super useful for testing purposes but probably not as useful for real world stuff. How does `importlib` manage this? It seems to me that it depends on `sys.path` and `sys.meta_path` for everything so it can't handle introspecting multiple environments without monkeypatching these. Presumably there are pypi packages that can inspect virtual environments that aren't active.

jaraco: I know `importlib.metadata.distributions()` takes an optional `path` argument, which defaults to `sys.path`, allowing a caller to inspect metadata for a set of paths. That could be pointed at a `site-packages` for a virtualenv to get the metadata for those packages, _but_ it won't honor other paths that might get added by the site config (such as a `venv --system-site-packages` or a `.pth` file in the site-packages that might extend the path). In my experience, the only technique I've seen for inspecting the true packaging environment for an interpreter is to launch that interpreter (in a subprocess) and then inspect that interpreter's environment.

That said, I know recent versions of `tox` have been doing some innovative things around detecting installed packages (and diffing them against an install spec), so there may be some good ideas in that codebase.

micropip/manager.py
ryanking13: `PackageManager` sounds good to me.

micropip/manager.py
ryanking13: > This is super useful for testing purposes but probably not as useful for real world stuff.

My main goal of making micropip modular is to allow us (or others) to customize some of its behavior. It's nice to have each instance behave independently, but it's not really something I'm very interested in right now (of course we can benefit from it in testing purpose).

ryanking13:
Thanks @RulerOfCakes!

hoodmane:
Thanks @RulerOfCakes!

Issue #88, opened 2023-09-27
Check the behavior of micropip.freeze with pyodide-lock


ryanking13:
`micropip.freeze` is now used in `pyodide-pack` (https://github.com/pyodide/pyodide-pack/pull/31), but I think [we don't have strong enough test cases for micropip.freeze](https://github.com/pyodide/micropip/blob/main/tests/test_freeze.py).

So I think it would be good to add some tests that validate the output of micropip.freeze with pyodide-lock.

PR #115, opened 2024-06-24
Port `freeze` into PackageManager

RulerOfCakes:
After #114, I'm continuing to implement each command into `PackageManager`.

As a relatively simple first change `freeze()` now uses the given state instead of the global variables.

Note that the current implementation of `freeze` still reads from `importlib.metadata.distributions()`, so it isn't completely independent from the global environment. This isn't as trivially separable as the `repodata_*` variables due to certain intrinsic quirks as discussed in the [previous PR](https://github.com/pyodide/micropip/pull/114#discussion_r1648020731), so for now I'm leaving the path/site-packages/install related elements as-is.


RulerOfCakes:
After the discussion about potentially having separate install directories in the previous PR, I've taken a look at [`tox`](https://github.com/tox-dev/tox) and from a quick overview it looks like their approach is to create an entirely separate python environment with `virtualenv` and avoid juggling with path / site-packages in the same environment. I'm unsure how to potentially incorporate this approach into `micropip` to work with `pyodide` as it doesn't seem to be an easy task, imo it feels like it would be better to leave it as a TODO for now. Any thoughts? @ryanking13

micropip/_commands/freeze.py
```diff
@@ -12,6 +12,12 @@
 
 
 def freeze() -> str:
+    return _freeze(REPODATA_PACKAGES, REPODATA_INFO)
+
+
+def _freeze(
+    repodata_packages: dict[str, dict[str, Any]], repodata_info: dict[str, str]
```
ryanking13, line 19: Let's change the parameter names to: `lockfile_packages`, `lockfile_info`.

micropip/_commands/freeze.py
```diff
@@ -23,18 +29,20 @@ def freeze() -> str:
...
     ``lockFileURL`` of :js:func:`~globalThis.loadPyodide`.
     """
-    return json.dumps(freeze_data())
+    return json.dumps(freeze_data(repodata_packages, repodata_info))
 
 
-def freeze_data() -> dict[str, Any]:
-    pyodide_packages = deepcopy(REPODATA_PACKAGES)
+def freeze_data(
+    repodata_packages: dict[str, dict[str, Any]], repodata_info: dict[str, str]
```
ryanking13, line 36: Here too.

micropip/package_manager.py
```diff
@@ -4,6 +4,7 @@
 )
 
 from micropip import package_index
+from micropip._commands.freeze import _freeze
```
ryanking13, line 7: I don't like `_commands` directory being imported from parent codes. So let's move `_freeze`, `freeze_data` functions to a separate file (e.g. `micropip/freeze.py`) and let both `_commands/freeze.py` and `packagee_manger.py` import from it.

tests/test_package_manager.py
```diff
@@ -27,3 +30,30 @@ def test_set_index_urls():
     finally:
         manager.set_index_urls(default_index_urls)
         assert manager.index_urls == default_index_urls
+
+
+def test_freeze():
+    manager = test_package_manager()
```
ryanking13, line 36:
```suggestion
    manager = PackageManager()
```

calling a test function inside another test function seems a bit weird.

ryanking13 (commented):
Thanks @RulerOfCakes!

ryanking13:
> Note that the current implementation of freeze still reads from importlib.metadata.distributions(), so it isn't completely independent from the global environment. This isn't as trivially separable as the repodata_* variables due to certain intrinsic quirks as discussed in the https://github.com/pyodide/micropip/pull/114#discussion_r1648020731, so for now I'm leaving the path/site-packages/install related elements as-is.

Yes, this can be quite complicated, so I think we can work on it separately later, for now you don't need to worry about making the PackageManager work independently.

micropip/_commands/freeze.py
ryanking13, line 19: (No action required for this PR)

For better typing and test coverage, we can think of using pyodide-lock package (which generates these repodata lock file) #88.

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
  from pyodide_lock import PackageSpec, InfoSpec
else:
  PackageSpec = ...
  InfoSpec = ...

_freeze(packages: PackageSpec, info: InfoSpec)
```

tests/test_package_manager.py
RulerOfCakes: True, I'll change `test_package_manager` to a non-test function.

tests/test_package_manager.py
```diff
@@ -27,3 +30,30 @@ def test_set_index_urls():
     finally:
         manager.set_index_urls(default_index_urls)
         assert manager.index_urls == default_index_urls
+
+
+def test_freeze():
+    manager = get_test_package_manager()
+
+    test_repodata_info = {
```
ryanking13, line 38: Later, I think we can improve this by using an actual lock file structure.

ryanking13 (approved):
Thanks @RulerOfCakes!

hoodmane:
Thanks @RulerOfCakes!

PR #119, opened 2024-07-06
Port `install` into `PackageManager`

RulerOfCakes:
- ported `install` into `PackageManager`
- moved implementation of `install` to parent directory(to avoid importing `_commands.install.py` from `package_manager.py`)

This concludes implementing the global commands in `PackageManager`, thus making it viable for use, although it doesn't really serve a separated purpose for end users at the moment. Adding customisability to inner behavior via this abstraction may be a later goal.


micropip/package_manager.py
```diff
@@ -6,31 +6,52 @@
...
 
     Each Manager instance holds its own local state that is
     independent of other instances.
-
-    TODO: Implement all of the following global commands to utilize local state.
     """
 
     def __init__(self) -> None:
-        self.index_urls = package_index.DEFAULT_INDEX_URLS
+        self.index_urls = package_index.DEFAULT_INDEX_URLS[:]
```
ryanking13, line 25: 👍

micropip/package_manager.py
```diff
@@ -6,31 +6,52 @@
 from micropip import package_index
 from micropip._commands import mock_package
 from micropip.freeze import freeze_lockfile
+from micropip.install import install
 from micropip.list import list_installed_packages
 from micropip.package import PackageDict
```
ryanking13, line 11: Let's change import paths to relative paths.

```
from .package import PackageDict
# same for other imports
```

ryanking13 (approved):
Thanks @RulerOfCakes!

micropip/package_manager.py
RulerOfCakes: Sure, but I'm curious about the reasoning behind this change? (Personally I don't see a significant advantage over one another)

ryanking13: It distinguishes the local imports from external imports. Also, a minor benefit would be that if we change the package name, it does not need to be updated.

ryanking13:
Thanks!

Issue #121, opened 2024-07-18
Should package index support "text/html; charset=utf-8" ?


Carreau:
I think that this line:

https://github.com/pyodide/micropip/blob/main/micropip/package_index.py#L227

Should likely be modified to:
```diff
-        case "application/vnd.pypi.simple.v1+html" | "text/html":
+        case "application/vnd.pypi.simple.v1+html" | "text/html" | "text/html; charset=utf-8":
            return partial(ProjectInfo.from_simple_html_api, pkgname=pkgname)
```

Or maybe be properly parsed ? because there might be more charset ?

It seem also that those kind of errors where reaching/parsing an index fails are swallowed (code behave as if the wheel was not found). I think the behavior should be stricter and if the index is wrong/malformatted unreachable we should errors and not behave as if the wheel was not found.

ryanking13:
> Or maybe be properly parsed

Sounds reasonable to me. Could you open a PR?

>  It seem also that those kind of errors where reaching/parsing an index fails are swallowed

Huh, it is not the behavior that I intended and I agree that it should show a more proper error message. I thought that the exception here:

https://github.com/pyodide/micropip/blob/5d4a8696f7daf302f060ffded34f9a84455fcffc/micropip/package_index.py#L230C13-L230C74

would be propagated to the user when the content type is not valid, but it doesn't work as expected, I guess?

Carreau:
Yeah, i need to re-dive into why it's not propagating exactly , but IIRC if it fails for any reason to find a package in the index  it falls back into looking in the lock file. Which make sense if the wheel is not found. 

I think that returning enums instead of raising in many places could help, but it would affect a large number of public APIs

Carreau:
This works now.

PR #124, opened 2024-07-21
Improve alternative index handling.

### LINKED ISSUE #123: Verbose setting overwrite custom logger level.
Transcations also have a verbose attribute which is unused.

`def install(..., verbose=<value default False>)`  overwrite a custom logger level. This and a few other places should allow `None` meaning "don't change the level".

Carreau:
This improve the handling of indexes and error messages.

Before this PR all errors reaching an index would be ValueError, thus not finding a wheel was the same as the index page having a non-handled content type, or even failing to parse the html because of a typo.

This made it hard to debug, or even realize that an index url was wrong, or that the pages served by the index were incorrect.

I thus introduce 3 New Errors that do not inherit from `ValueError`:
 - IndexMetadataFetchError
 - UnsupportedParserContentTypeError
 - WheelNotFoundError

The first two of which trigger real error, as they likely suggest a problem with the index or the user config, and should not be ignored.

In addition the `verbose=` option was unconditionally setting the log level. I think this is improper as if the logging level is set somewhere else (to DEBUG, or something else), then it is overwritten.

- This then adds the option to pass `verbose=None`, (and make it the default), in which case it will not change the default.

python -m http.server will serve ContentType with `; charset=utf-8`, which is not recognized.

- This now handle the case where the index ContentType header contains parameters for example `; charset=utf-8`, by discarding everything after the semicolon `;`; this is not proper parsing, but should be sufficient here.

 - I found that more debug logging would be useful and added a number of debug logs calls

With this I appear to get proper error message and debug statements when trying to work on the Cors proxy.

Should close #123, and #121, and help with pyodide/pyodide#4898

### DIFF
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -269,6 +269,10 @@ class IndexMetadataFetchError(BaseException):
     pass
 
 
+class IndexMetadataParseError(BaseException):
+    pass
+
+
 async def query_package(
     name: str,
     fetch_kwargs: dict[str, Any] | None = None,
@@ -318,6 +322,8 @@ async def query_package(
         try:
             metadata, headers = await fetch_string_and_headers(url, _fetch_kwargs)
         except OSError as e:
+            # Pypi does not currently set cors on 404, we we can't know
+            # whether this is a real OS error or a 404
             logger.debug("Error fetching %r, skipping. Error was %r", url, e)
             continue
 
@@ -326,7 +332,9 @@ async def query_package(
         try:
             return parser(metadata)
         except ValueError as e:
-            raise IndexMetadataFetchError("Error parsing Index page") from e
+            raise IndexMetadataParseError(
+                f"Error parsing Index page for URL: {url} "
+            ) from e
     else:
         raise IndexMetadataFetchError(
             f"Can't fetch metadata for '{name}'. "
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -322,7 +322,7 @@ async def query_package(
         try:
             metadata, headers = await fetch_string_and_headers(url, _fetch_kwargs)
         except OSError as e:
-            # Pypi does not currently set cors on 404, we we can't know
+            # PyPI does not currently set cors on 404, we can't know
             # whether this is a real OS error or a 404
             logger.debug("Error fetching %r, skipping. Error was %r", url, e)
             continue
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import string
 import sys
 from collections import defaultdict
@@ -19,6 +20,8 @@ DEFAULT_INDEX_URLS = ["https://pypi.org/simple"]
 INDEX_URLS = DEFAULT_INDEX_URLS
 
 _formatter = string.Formatter()
+
+logger = logging.getLogger("micropip")
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


Carreau:
There are still a few case that needs handling, in particular when fetching for a given wheel and the page does not exists, is it a problem with the index, or does the package really not exists. 

I think I might have to tweak the behavior a bit more.

Carreau:
Ok, one of the issues is that in particular PYPI does not set CORS on 404, therefore we get an OS error when trying to fetch a non-existing page. 

I'm wondering if this is a bug upstream

hoodmane:
Thanks @Carreau.

> PYPI does not set CORS on 404

This does seem to be a warehouse bug.

Carreau:
> Thanks @Carreau.
> 
> > PYPI does not set CORS on 404
> 
> This does seem to be a warehouse bug.

Yep and it's known  https://github.com/pypi/warehouse/issues/14229

micropip/package_index.py
```diff
@@ -273,17 +317,26 @@ async def query_package(
             url = url.format(package_name=name)
         else:
             url = f"{url}/{name}/"
+        logger.debug("Searching index url %r", url)
 
         try:
             metadata, headers = await fetch_string_and_headers(url, _fetch_kwargs)
-        except OSError:
+        except OSError as e:
+            # Pypi does not currently set cors on 404, we we can't know
```
ryanking13, line 325:
```suggestion
            # PyPI does not currently set cors on 404, we can't know
```

ryanking13 (approved):
Thanks. Looks good to me overall. Could you update [the changelog](https://github.com/pyodide/micropip/blob/main/CHANGELOG.md) and fix the failing tests? I think you need to update the Error type [here](https://github.com/pyodide/micropip/blob/5d4a8696f7daf302f060ffded34f9a84455fcffc/tests/test_package_index.py#L136)

Carreau:
Yeah, I'm looking at understanding the failing tests.

I'd like to figure out how to both get proper error messages if the index is invalid, and just keep going on 404, but because of PyPI not setting CORS on 404 I don't know if it's currently possible. 

So I'm wondering if i I should bake a an boolean for indexes `no_cors_on_404` or similar. Otherwise we'll get back to original behavior of swallowing almost every errors (including parse error). I'm not even sure the case where an index return a real 404 with proper cors is properly handled, and does not raise a parse error as query_package only error handing is OSError and does not check response status code, or does not use raises_for_status.

(I also have limited time, so might not touch this that much by next week).

hoodmane:
We should look into fixing the warehouse bug. I tried a little bit but I didn't get the warehouse test suite running.

ryanking13:
Yep, +1 for fixing the warehouse bug. Let me see if I can understand and modify the warehouse codebase.

Carreau:
Also https://github.com/pyodide/pyodide/pull/4974 will be relevant for later refactor.

Carreau:
crosslinking https://github.com/pypi/warehouse/pull/16339 as well.

Carreau:
So should I assume that we'll always get micropip on the latest pyodide, or can we have conditional that depends on versions ?

ryanking13:
If there are some changes in Pyodide and if you want to use those features in the micropip, you are welcome to do so.

We try to keep the compatibility of micropip and Pyodide versions if possible, but it is not mandatory.

Carreau:
I'm going to close for now, there are too many conflicts, but I may coma back to it later.

PR #125, opened 2024-07-23
Allow injecting kwargs to fetch during installing

### LINKED ISSUE #122: Maybe provide a parameter to avoid browser caching for PyPI resolution requests
Currently micropip passes a kwargs to `pyfetch` during resolution.

https://github.com/pyodide/micropip/blob/5d4a8696f7daf302f060ffded34f9a84455fcffc/micropip/install.py#L138

I think it is reasonable to add a `no_cache` parameter to `install` command or the `Transaction` class to skip browser caching. Currently it is not possible to skip browser caching for PyPI simple API calls.

CNSeniorious000:
Resolves #122 

But will the next version released after implementing #112? If this feature is going to be added there then there is no need to merge this tiny commit.

### DIFF
--- a/micropip/install.py
+++ b/micropip/install.py
@@ -20,6 +20,7 @@ async def install(
     index_urls: list[str] | str | None = None,
     *,
     verbose: bool | int = False,
+    **fetch_kwargs,
 ) -> None:
     """Install the given package and all of its dependencies.
 
@@ -113,8 +114,6 @@ async def install(
     ctx = default_environment()
     if isinstance(requirements, str):
         requirements = [requirements]
-
-    fetch_kwargs = dict()
 
     if credentials:
         fetch_kwargs["credentials"] = credentials


micropip/install.py
```diff
@@ -20,6 +20,7 @@ async def install(
     index_urls: list[str] | str | None = None,
     *,
     verbose: bool | int = False,
+    **fetch_kwargs,
```
ryanking13, line 23: Let's not use kwargs, and pass a dictionary instead.

``
fetch_kwargs: dict[str, Any] | None = {...}
``

ryanking13 (commented):
Thanks! this would need some test and changelog update.

> But will the next version released after implementing https://github.com/pyodide/micropip/issues/112?

I don't think so, #112 will take more time to be fully implemented and released. And even after #112 is implemented I think this PR is still relavant.

micropip/install.py
ryanking13: Also, as `credentials` is part of fetch args, I think we can deprecate it in favor to fetch args.

PR #126, opened 2024-07-23
Use parametrize test instead of fixture.

Carreau:
This is useful to run only a subset of test with --ff or --lf when trying to debug/fix a situation and should make further parametrisation a bit more easy.

query_package is used only in a few location so I woudl also be tempted to make it only accept lists instead os str|list[str] but that's another discussion.

We could further parametrize on

     for _index_urls in (
         pkg1_index_url,
         [pkg1_index_url],
         [pkg2_index_url, pkg1_index_url],

But it woudl be a bit more complicated.


Carreau:
Thanks !

PR #127, opened 2024-07-25
Organizing the compatibility layer

RulerOfCakes:
- Reorganized _compat.py into a separate directory.
- Added `CompatibilityLayer` as a base class structure to be implemented by both environments(inside/outside of pyodide)


ryanking13 (approved):
Thanks! Looks good to me.

So, I guess we could inject this compatibility layer to micropip object so users can modify the behavior of micropip, right?

ryanking13:
FYI, I am planning to reduce the number of compatibility functions that micropip should import from pyodide by abstracting the package installation (https://github.com/pyodide/pyodide/issues/4914)

RulerOfCakes:
> Thanks! Looks good to me.
> 
> So, I guess we could inject this compatibility layer to micropip object so users can modify the behavior of micropip, right?

Yes, but I don't think it necessarily has to be injected into the `PackageManager` object as it it is being used more like an external library in the codebase. Unless there would also be needs to modify these exported functions as well?

ryanking13:
> Unless there would also be needs to modify these exported functions as well?

Maybe in the future, we may provide such a feature, but yes, it is not our goal right now.

Thanks again!

hoodmane:
Thanks @RulerOfCakes!

PR #128, opened 2024-07-26
DOCS Set html_baseurl

hoodmane:
readthedocs used to set this automatically but they say they've deprecated that behavior.


ryanking13:
Thanks. I also got the mail from RTD I haven't read it properly yet. Thanks for taking care of this.

PR #129, opened 2024-08-06
BUG: fetch_string_and_headers compat: raise in and out of pyodide

Carreau:
Currently only the not_in_pyodide will raise on non-success, because this is the default behavior of urllib, the in_pyodide will not, so I added a raise_for_status.

It is better to raise, as otherwise the package parser will potentially get proper URL and not manage to parse it, and decide there is no wheels, while we actually just got an error (404, or maybe 500).

In addition wraps both case in a custom local HttpStatusError, so that we can actually catch these errors in the right places when we encounter them.

I think we could define HttpStatusError on the ABC and not make it an ABC itself, but it small enough that I think duplication is ok.

### DIFF
--- a/micropip/_compat/_compat_in_pyodide.py
+++ b/micropip/_compat/_compat_in_pyodide.py
@@ -5,9 +5,21 @@ from urllib.parse import urlparse
 if TYPE_CHECKING:
     pass
 
+import pyodide
+from packaging.version import parse
 from pyodide._package_loader import get_dynlibs
 from pyodide.ffi import IN_BROWSER, to_js
-from pyodide.http import pyfetch
+
+if parse(pyodide.__version__) > parse("0.27"):
+    from pyodide.http import HttpStatusError, pyfetch
+else:
+
+    class HttpStatusError(Exception):  # type: ignore [no-redef]
+        """we just want this to be defined, it is never going to be raised"""
+
+        pass
+
+    from pyodide.http import pyfetch
 
 from .compatibility_layer import CompatibilityLayer
 
@@ -28,6 +40,15 @@ except ImportError:
 
 
 class CompatibilityInPyodide(CompatibilityLayer):
+    class HttpStatusError(Exception):
+        status_code: int
+        message: str
+
+        def __init__(self, status_code: int, message: str):
+            self.status_code = status_code
+            self.message = message
+            super().__init__(message)
+
     @staticmethod
     def repodata_info() -> dict[str, str]:
         return REPODATA_INFO
@@ -50,7 +71,11 @@ class CompatibilityInPyodide(CompatibilityLayer):
     async def fetch_string_and_headers(
         url: str, kwargs: dict[str, str]
     ) -> tuple[str, dict[str, str]]:
-        response = await pyfetch(url, **kwargs)
+        try:
+            response = await pyfetch(url, **kwargs)
+            response.raise_for_status()
+        except HttpStatusError as e:
+            raise CompatibilityInPyodide.HttpStatusError(e.status, str(e)) from e
 
         content = await response.string()
         headers: dict[str, str] = response.headers
--- a/micropip/_compat/_compat_in_pyodide.py
+++ b/micropip/_compat/_compat_in_pyodide.py
@@ -5,21 +5,9 @@ from urllib.parse import urlparse
 if TYPE_CHECKING:
     pass
 
-import pyodide
-from packaging.version import parse
 from pyodide._package_loader import get_dynlibs
 from pyodide.ffi import IN_BROWSER, to_js
-
-if parse(pyodide.__version__) > parse("0.27"):
-    from pyodide.http import HttpStatusError, pyfetch
-else:
-
-    class HttpStatusError(Exception):  # type: ignore [no-redef]
-        """we just want this to be defined, it is never going to be raised"""
-
-        pass
-
-    from pyodide.http import pyfetch
+from pyodide.http import HttpStatusError, pyfetch
 
 from .compatibility_layer import CompatibilityLayer
 
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -280,23 +280,6 @@ async def query_package(
             if e.status_code == 404:
                 continue
             raise
-        except OSError:
-            # temporary pyodide compatibility.
-            # pypi now set proper CORS on 404 (https://github.com/pypi/warehouse/pull/16339),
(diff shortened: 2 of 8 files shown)


tests/test_compat.py
```diff
@@ -0,0 +1,17 @@
...
+test that function in compati behave the same
+
+"""
+
+import pytest
+
+from micropip._compat import HttpStatusError, fetch_string_and_headers
+
+
+@pytest.mark.asyncio
```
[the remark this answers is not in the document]
hoodmane: It's a bit more complicated. Honestly, the easiest thing might be just to copy paste the code. At some point we'll have to think about a better way...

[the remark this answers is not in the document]
ryanking13: Yep, testing something both in and out of Pyodide is a bit complex. Actually, I think you don't need to test out-of-pyodide case. We don't care about using micropip out of Pyodide and `not_in_pyodide` exists mostly for test purpose.

Carreau:
Moving to draft, I think HttpStatusError is not yeat in latest released pyodide.

Carreau:
Ok, I've added relevant fixes now that PyPI properly set CORS on 404s, and added conditional for wether we are on pyodide 0.26.x or 0.27+.

Carreau:
SO now some of the fixture that return 500 now need to return 404... I'm not sure how to get the httpserver fixture to return 404 on wildcard url without expecting requests.

ryanking13:
> ValueError: Function test_404_raise's first argument name 'httpserver' should start with 'selenium'

`@run_in_pyodide` should be the inner-most decorator.

tests/test_compat.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+
+import pytest
+from pytest_pyodide import run_in_pyodide
+
+from micropip._compat import HttpStatusError, fetch_string_and_headers
+
+
+@run_in_pyodide
+@pytest.mark.asyncio
```
ryanking13, line 13:
```suggestion
@pytest.mark.asyncio
@run_in_pyodide
```

ryanking13:
Let's release Pyodide 0.27.0a1 and use it in the micropip CI so we don't need to put complex version checks. Hood released 0.27.0a1 yesterday but the deploy was failed because of the CI timeout (https://github.com/pyodide/pyodide/commit/5231ac0358cf3787c223d699bd36e8648de43792). Let me see if rerunning the job would work.

Carreau:
Ok, I've tried multiple variation of the test to `@run_in_pyodide`, but can't figure out how to actually have one that works.

tests/test_compat.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+
+import pytest
+from pytest_pyodide import run_in_pyodide
+
+from micropip._compat import HttpStatusError, fetch_string_and_headers
+
+
+@run_in_pyodide
+async def _inner_test_404_raise(httpserver):
```
hoodmane, line 13:
```suggestion
async def _inner_test_404_raise(selenium, httpserver):
```

hoodmane: Also the `httpserver` fixture isn't going to work with `@run_in_pyodide`. I'm not sure what the best way is to do this, @ryanking13 do you have any suggestions?

hoodmane:
I released 0.27.0a2 instead.

tests/test_compat.py
ryanking13: You can do this way (passing only string not httpserver fixture)

```py
def test_404(httpserver):
    @run_in_pyodide
    async def _inner_test_404_raise(url):
        from ... import fetch_string_and_headers
        import pytest
        with pytest.raises(HttpStatusError):
            await fetch_string_and_headers(url, {})

    httpserver.expect_request("/404").respond_with_data(
        "Not found", status=404, content_type="text/plain"
    )
    url_404 = httpserver.url_for("/404")
    _inner_test_404_raise(url_404)
```

Carreau: Ah... let me try that, thanks for the pointers.

Carreau:
> I released 0.27.0a2 instead.

BTW should CI be updated 

```diff
- pyodide-version: [0.24.1, 0.25.0]
+ pyodide-version: [0.24.1, 0.25.0, 0.26.0, 0.27.0a2]
```

?

ryanking13:
>  Any idea on how to actually change the timeout to something else the 10 minutes... I've tried various options with no success.

I think this is a chrome (+selenium) issue.. for now I just pinned the version of the chrome to some older version to see if it pass. https://github.com/pyodide/micropip/pull/135

Carreau:
> I think this is a chrome (+selenium) issue.. for now I just pinned the version of the chrome to some older version to see if it pass. #135

More generally when I write a test with an error the timeout is also 10 min, and I could not find how to make it timeout in less... but thanks for #135

tests/conftest.py
```diff
@@ -72,6 +72,7 @@ def selenium_standalone_micropip(selenium_standalone, wheel_path):
     if not wheel_files:
         pytest.exit("No wheel files found in wheel/ directory")
 
+    # Change  default timeout to 30s instead of 10min
```
Carreau, line 75:
```suggestion
```

micropip/_compat/_compat_in_pyodide.py
```diff
@@ -5,9 +5,21 @@
...
-from pyodide.http import pyfetch
+
+if parse(pyodide.__version__) > parse("0.27"):
+    from pyodide.http import HttpStatusError, pyfetch
+else:
+
+    class HttpStatusError(Exception):  # type: ignore [no-redef]
+        """we just want this to be defined, it is never going to be raised"""
+
+        pass
```
ryanking13, line 20: I removed Pyodide versions < 0.27 from CI. I am fine with removing these conditions and drop backward compatibility for versions < 0.27 @hoodmane WDYT?

micropip/package_index.py
```diff
@@ -276,11 +276,33 @@ async def query_package(
...
+                import pyodide
+                from packaging.version import parse
+
+                if parse(pyodide.__version__) > parse("0.27"):
+                    # reraise on more recent pyodide.
+                    raise
+                continue
+            except ImportError:
+                # not in pyodide.
+                raise
```
ryanking13, line 299: Here too.

Carreau:
ok, I pushed a commit that remove compat with old versions.

ryanking13 (approved):
Thanks @Carreau!

PR #130, opened 2024-08-26
Bump Pyodide on CI

Carreau:
Also change to explicitely use strings.

I'm guessing this should be bumped, maybe there is an automatic way to do latest ? Also not sure if you want to test on alphas, but I guess it's always useful to catch bugs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        pyodide-version: [0.24.1, 0.25.0]
+        pyodide-version: ['0.24.1', '0.25.0', '0.26.2', '0.27.0a2']
         test-config: [
           {runner: selenium, runtime: chrome, runtime-version: latest },
         ]


tests/test_uninstall.py
```diff
@@ -188,10 +191,12 @@ async def run(selenium, pkg_name, wheel_url):
...
-
-            assert len(logs) == 2
-            assert "does not exist" in logs[-1]
-            assert "does not exist" in logs[-2]
+            if parse(pyodide.__version__) < parse("0.27"):
+                assert len(logs) == 2, logs
+                assert "does not exist" in logs[-1]
+                assert "does not exist" in logs[-2]
+            else:
+                assert logs == [""]
```
hoodmane, line 199: @ryanking13 Looks like a regression?

hoodmane:
Thanks @Carreau! Why not split this into a PR which adds 0.26 which we certainly want, and a second PR that adds 0.27.0a2 so we can investigate the failure there and make an appropriate fix without blocking the addition of 0.26?

Carreau:
0.26 also has the regression, and there does not seem to be any other regression between 0.26 and 0.27.0a2, so hopefully it's not more difficult.  I also see selenium Timeouts and that I'm a bit flumoxed.

Carreau:
Also any pointers on change the value of the selenium timeouts ? I think it's 600 for now, (i'm guessing seconds ?). That would help making failing test faster

Carreau:
Thanks !

PR #131, opened 2024-08-26
Do not use f-strings in log calls.

Carreau:
It is in general bad practice; in particular if one want a structured logging handler that store the template string and the values separately, using f-strings makes it not possible.

I belive there is also a question of actually calling the formatter on the interpolated values which is not done if the loging level is low enough.

WHile it might not be too critical for micropip, I belive it is good practive to show the example.

### DIFF
--- a/micropip/_commands/uninstall.py
+++ b/micropip/_commands/uninstall.py
@@ -39,7 +39,7 @@ def uninstall(packages: str | list[str], *, verbose: bool | int = False) -> None
             dist = importlib.metadata.distribution(package)
             distributions.append(dist)
         except importlib.metadata.PackageNotFoundError:
-            logger.warning(f"Skipping '{package}' as it is not installed.")
+            logger.warning("Skipping 's' as it is not installed.", package)
 
     for dist in distributions:
         # Note: this value needs to be retrieved before removing files, as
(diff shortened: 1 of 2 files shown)


Carreau:
I guess in a couple of places, `%r`, is better suited than `'%s'`, but I didn't want to be too invasive with two semantically different changes that might make review difficult.

Carreau:
See among other all the https://docs.astral.sh/ruff/rules/#flake8-logging-format-g

Carreau:
Actually, if there is a ruff lint, why not add it.

ryanking13 (approved):
Thanks @Carreau! I added all those logging stuff, and to be honest, I don't know much about the good logging practice. So please feel free to make suggestions if you found any issues with logging!

Carreau:
I think this is perfectly fine for a project like micropip, but I like to show good practice. 

Just think that for really huge webapp, like facebook level, you want to store the template string and the values separately (potentially you store the template string only once), and not ins a log file but with mysql with 1 column per value.  So that you can ask "show me all the identical messages", and/or if one of the value is say a float like in 
`log("page took %s ms to render", time)`, you can ask for mean/std. 

And with that all those "don't use f-string" in logs will make sens.

I don't personally do use those kind of infra, but just knowing the above made it make sens to me.

ryanking13:
The failing test isn't related to this PR (timeout in the chrome browser). Let me go ahead and merge.

Carreau:
> The failing test isn't related to this PR (timeout in the chrome browser). Let me go ahead and merge.

Thanks, I keep seeing those timeout, and don't know what to do about those.

PR #132, opened 2024-08-26
Has setup_logging not override already setup loging level.

Carreau:
Add a verbosity=None (and make it the default) for verbosity. If one has already increase the logging level, the default value of verbose should not change it. This is achieved with a default value of None.

Technically I guess we should have a context manager to restaure previous value if True/False is passed, but that will be for another time.

I also think that if verbosity is an int, we should set the lgging level to that int instead of havin an arbitrary verbosity of 0/1/2; but that's also another discussion.

I'm doing this as I'd like adding some logger.debug, but I don't want to pass verbose=True everywhere.

### DIFF
--- a/micropip/logging.py
+++ b/micropip/logging.py
@@ -98,7 +98,7 @@ class LoggerWrapper:
     def __getattr__(self, attr):
         return getattr(self.logger, attr)
 
-    def __setattr(self, attr, value):
+    def __setattr__(self, attr, value):
         return setattr(self.logger, attr, value)
 
     @contextmanager
--- a/micropip/logging.py
+++ b/micropip/logging.py
@@ -1,14 +1,14 @@
-import contextlib
 import logging
 import sys
 from collections.abc import Generator
+from contextlib import contextmanager
 from typing import Any
 
 _logger: logging.Logger | None = None
 _indentation: int = 0
 
 
-@contextlib.contextmanager
+@contextmanager
 def indent_log(num: int = 2) -> Generator[None, None, None]:
     """
     A context manager which will cause the log output to be indented for any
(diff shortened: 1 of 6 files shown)


micropip/logging.py
```diff
@@ -87,17 +87,16 @@ def _set_formatter_once() -> None:
...
-    _logger.setLevel(level_number)
-
+    if verbosity is not None:
+        if verbosity >= 2:
+            level_number = logging.DEBUG
+        elif verbosity == 1:  # True == 1
+            level_number = logging.INFO
+        else:
+            level_number = logging.WARNING
+        print("Will set log level to:", level_number)
```
hoodmane, line 100: Do we really want to print here?

Carreau:
```suggestion
```

hoodmane:
I'm not sure I like the fact that this makes the `logging` state use the level from the previous call. I think maybe `setup_logging` could return a manager object with methods to restore the log level at the end of the function and around `await`. And add a `set_log_level` method. Something like this:
https://github.com/pyodide/micropip/commit/b877ee2829aa5949c66150283449bebba6e29277

WDYT?

Carreau:
Yeah, if you want to push further improvement I'm all for it. feel free to push on my branch.

Alternative is to have set_log_level unconditionally a context manager.

hoodmane:
Okay I pushed my changes here.

hoodmane:
So now I think we should expose `micropip.set_log_level` and deprecate the logging argument to `micopip.install` and `micropip.uninstall`.

Carreau:
> argument to `micropip.install` and `micropip.uninstall`

I see how those can be useful as long as they temporarily increase/decrease level and the default is to no change. IMHO verbose=None|False|True is ment for end user, while setting to level like `DEBUG` is really for developers. 

Regardless it's up to you.

micropip/logging.py
```diff
@@ -87,17 +87,48 @@ def _set_formatter_once() -> None:
...
+        try:
+            yield
+        finally:
+            self.logger.setLevel(cur_level)
+
 
+def setup_logging(verbosity: int | bool | None) -> LoggerWrapper:
+    _set_formatter_once()
+    assert _logger
+    result = LoggerWrapper(_logger)
```
Carreau, line 113: I think this is still wrong, as this changes the  logging level permanently for a logger wrapper even if passed None. 

I think the conflation of getting a logger and setting the level is creating issues.

Carreau:
Thanks for your patience, and sorry I was on PTO.

Back at this again; 

as setup_logging is used only in 2 places I think it's ok to have a bigger change on API. 

`setup_logging()` now takes no arguments, and now return a loggerwrapper with `.level(None|True|False|0|1|2)` context manager that restore to previous level on exit.


All two usages are now using this new API, this so install and uninstall should behave the same, except if the level is changed outside of those calls, it stay changed.

Carreau:
Failure seem unrelated.

micropip/logging.py
```diff
@@ -87,17 +87,56 @@ def _set_formatter_once() -> None:
...
+
+
+def setup_logging() -> LoggerWrapper:
     _set_formatter_once()
+    assert _logger
+    return LoggerWrapper(_logger)
+
 
+# TODO: expose this somehow
+def set_log_level(verbosity: int | bool):
```
ryanking13, line 134: Do we still need this function?

micropip/logging.py
```diff
@@ -87,17 +87,56 @@ def _set_formatter_once() -> None:
...
+
+    logger: logging.Logger
+
+    def __init__(self, logger: logging.Logger):
+        self.logger = logger
+
+    def __getattr__(self, attr):
+        return getattr(self.logger, attr)
+
+    def __setattr(self, attr, value):
```
ryanking13, line 101: Shouldn't this be `__setattr__`?

ryanking13 (approved):
Thanks for improving the logger @Carreau! It looks good to me and the CI failure is not related.

micropip/logging.py
Carreau: probably not...

Carreau:
Thanks !

PR #137, opened 2024-09-18
Maint: Show test duration and limit to 15 failures.

Carreau:
I don't think it is worth contiuing testing if there is more than 15 failures.

Also show the 10 slowest test, I'm hopping that will give us insight into how to make testing faster.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,9 @@ jobs:
         run: |
           pytest -v \
             --cov=micropip \
+            --durations=10 \
             --dist-dir=./dist/ \
+            --maxfail=15 \
             --runner=${{ matrix.test-config.runner }} \
             --rt ${{ matrix.test-config.runtime }}
 


micropip/logging.py
```diff
@@ -88,12 +88,13 @@ def _set_formatter_once() -> None:
...
-    __slots__ = ("logger", "_orig_level")
 
-    logger: logging.Logger
+    # need a default value because of __getattr__/__setattr__
+    logger: logging.Logger = None  # type: ignore[assignment]
 
     def __init__(self, logger: logging.Logger):
-        self.logger = logger
+        # Bypassing __setattr__ by setting attributes directly in __dict__
+        self.__dict__["logger"] = logger
```
hoodmane, line 97: What's the point of this? I don't understand.

.github/workflows/main.yml
```diff
@@ -52,7 +52,9 @@ jobs:
         run: |
           pytest -v \
             --cov=micropip \
+            --durations=10 \
             --dist-dir=./dist/ \
+            --maxfail=15 \
```
hoodmane, line 57: Looks good thanks!

micropip/logging.py
Carreau: `self.logger = ...` call setattr, which itself, need  `self.logger`, (in `getattr(self.logger, ...,...)`, though `self.logger` is not defined so goes through getattr (infinite recursion). OTOH `self.__dict__` exists, so all fine. 

Basically #138 should be merged first, it fixes the failure on main, and then I'll rebased my own PRs on top of #138.

Carreau:
Rebased to remove the commits of #138 now that it is merged.

agriyakhetarpal (approved):

agriyakhetarpal:
Thanks, @Carreau!

PR #138, opened 2024-09-18
Fix: infinite recursion on LoggerWrapper

Carreau:


Carreau:
Not sure why it was not caught on the original PR test suite.

ryanking13 (approved):
Thanks!

PR #139, opened 2024-09-19
Support PEP-658

ryanking13:
This PR adds [PEP 658: Serve Distribution Metadata in the Simple Repository API](https://peps.python.org/pep-0658/) support to micropip.

When downloading package using Simple API (which is not a default), micropip will now check for distribution metadata and download the metadata file simultaneously which (hopefully) speeds up the dependency resolution.


micropip/wheelinfo.py
```diff
@@ -165,7 +196,7 @@ async def _fetch_bytes(self, fetch_kwargs: dict[str, Any]):
                 raise e
             else:
                 raise ValueError(
-                    f"Can't fetch wheel from '{self.url}'. "
+                    f"Can't fetch wheel from '{url}'. "
```
Carreau, line 199:
```suggestion
                    f"Can't fetch wheel from {url!r}. "
```

Carreau (approved):
Personally I don't like adding gz files to repositories as this is the way [XZ backdoor](https://en.wikipedia.org/wiki/XZ_Utils_backdoor#Mechanism) worked.

ryanking13:
> Personally I don't like adding gz files to repositories as this is the way [XZ backdoor](https://en.wikipedia.org/wiki/XZ_Utils_backdoor#Mechanism) worked.

Yeah, we've removed gz files in this repository so I think I should remove it in this PR too.

ryanking13:
@hoodmane @agriyakhetarpal Can one of you review this when you have time? I would like to include this in the next release.

micropip/transaction.py
```diff
@@ -247,8 +247,16 @@ async def add_wheel(
...
         if self.deps:
+            try:
+                await wheel.download_pep658_metadata(self.fetch_kwargs)
+            except OSError:
+                # If something goes wrong while downloading the metadata
+                # we just log the error and continue, as the metadata can be fetched extracted from the wheel.
+                logger.debug("Metadata not available for %s", wheel.name)
+
+            await wheel_download_task
             await self.gather_requirements(wheel.requires(extras))
```
hoodmane, line 260: If `download_pep658_metadata()` succeeds is it possible to exchange the order here and call `gather_requirements()` before awaiting `wheel_download_task`?

ryanking13:
> If we can resolve the dependencies before the wheel arrives hopefully that's a performance win.

Yes, that's what I hope to achieve. How much it actually helps performance will depend on the network condition though.

agriyakhetarpal:
> @hoodmane @agriyakhetarpal Can one of you review this when you have time? I would like to include this in the next release.

Thanks for the ping! Yes, I can help review it too. I am not acquainted with this PEP, so I'll need to read a bit about it.

micropip/transaction.py
ryanking13: Oh, good catch. I originally implemented in that way in #90... probably I did something wrong while cherry-picking changes.

agriyakhetarpal (commented):
I've read the PEP now; these changes mostly look good to me. Thank you, @ryanking13! I hope this will give some beneficial resolution speedups before the full wheel downloads kick in:

However, I think this PEP should be accompanied by [PEP 714](https://peps.python.org/pep-0714/#specification) as well, which fixed a few bugs in PEP 658's implementation because PyPI and `pip` both had bugs that caused the metadata not to be parsed correctly. While I am confident that we haven't introduced that bug here, we should remain closer to PyPI regarding the implementation, so we should account for that change here. The key changes are as follows (please let me know if I missed something):
- for the HTML API, rename to `data-core-metadata`, and for the JSON API, rename to `core-metadata`, and
- fall back to the currently proposed `data-dist-info-metadata` if they don't exist

Details: what I understand from that PEP is that PyPI kept the current/broken keys but also added these new keys for backward compatibility, from this line in the PEP:

> To support clients that used the previous key names, the HTML representation **MAY** also be emitted using the `data-dist-info-metadata`, and if it does so it **MUST** match the value of `data-core-metadata`.

And for the changes to add to `micropip` (client-side):

> Clients consuming the JSON representation of the Simple API **MUST** read the [PEP 658](https://peps.python.org/pep-0658/) metadata from the key `core-metadata` if it is present. They **MAY** optionally use the legacy `dist-info-metadata` key if it is present but `core-metadata` is not.

I don't intend to block the merge in case the release is urgently needed (apologies for the delay in the review); but I do feel that we should incorporate both PEPs in the same PR, since I don't think a lot of further changes would be required; just a few renames to the new APIs added, some fallbacks, and a few changes to the testing indices (which I am not familiar with the code for, though).

Here is `pip`'s fix: pypa/pip#12078, and similarly for Warehouse: pypi/warehouse#13706

<hr>

Also, one slightly related point for `self.parsed_url.hostname`: the Anaconda.org index does not seem to support this PEP on the server side, which is an aspect that we should put into some consideration when we switch to using `https://anaconda.org/pyodide/simple` in pyodide/pyodide-build#43 (while that is still okay because it's just for the build-time dependencies) and if we start using it for https://github.com/pyodide/pyodide-recipes later on to host most (if not all) wheels post pyodide/pyodide#4918:

`pip` from PyPI:

<details><summary>Tap to expand</summary>
<p>

```
$ pip install cython --dry-run --no-cache
```

```
Collecting cython
  Downloading Cython-3.0.11-py2.py3-none-any.whl.metadata (3.2 kB)
Downloading Cython-3.0.11-py2.py3-none-any.whl (1.2 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.2/1.2 MB 37.1 MB/s eta 0:00:00
Would install Cython-3.0.11
```

</p>
</details> 

However, when downloading from Anaconda.org:

<details><summary>Tap to expand</summary>
<p>

```
$ pip install -i https://pypi.anaconda.org/pyodide/simple cython --dry-run --no-cache
```

```
Looking in indexes: https://pypi.anaconda.org/pyodide/simple
Collecting cython
  Downloading https://pypi.anaconda.org/pyodide/simple/cython/3.0.11/Cython-3.0.11-py2.py3-none-any.whl (1.2 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.2/1.2 MB 1.8 MB/s eta 0:00:00
Would install Cython-3.0.11 
```

</p>
</details>

<hr>

Unrelated point, but similar to @Carreau's suggestion above: we should remove the existing wheel files from `tests/test_data/wheel/`, since they will only bloat the repository as we proceed to add more commits to the default branch. They can be stored elsewhere (maybe in a GitHub release in a new repository and downloaded into the directory using `pooch` before running the test suite).

ryanking13:
> However, I think this PEP should be accompanied by [PEP 714](https://peps.python.org/pep-0714/#specification) as well, which fixed a few bugs in PEP 658's implementation because PyPI and pip both had bugs that caused the metadata not to be parsed correctly.

Oh, that's a good point indeed. I think updating the mousevendor would work, so let me update the PR.

ryanking13:
> Unrelated point, but similar to @Carreau's suggestion above: we should remove the existing wheel files from tests/test_data/wheel/, since they will only bloat the repository as we proceed to add more commits to the default branch. They can be stored elsewhere

Yes, maybe let's discuss it in a separate issue. I don't like to put bunch of big files in the repository for testing, but I am also not sure whether putting it in a separate place is a good way to go.

micropip/wheelinfo.py
```diff
@@ -130,10 +135,42 @@ async def download(self, fetch_kwargs: dict[str, Any]):
...
+        self._data = await self._fetch_bytes(self.url, fetch_kwargs)
+
+        # The wheel's metadata might be downloaded separately from the wheel itself.
+        # If it is not downloaded yet or if the metadata is not available, extract it from the wheel.
+        if self._metadata is None:
+            with zipfile.ZipFile(io.BytesIO(self._data)) as zf:
+                metadata_path = (
+                    wheel_dist_info_dir(zf, self.name) + "/" + Metadata.PKG_INFO
+                )
+                self._metadata = Metadata(zipfile.Path(zf, metadata_path))
```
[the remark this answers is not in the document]
ryanking13: Sounds good

ryanking13:
>  Also, the current tests/test_data/pypi_response/pytest_simple.html does contain data-core-metadata and not data-dist-info-metadata, so we should indeed account for the change in micropip/externals/mousebender/simple.py.

I just checked the test file and it seems like it contains both `data-core-metadata` and `data-dist-info-metadata`.

agriyakhetarpal:
> > Also, the current tests/test_data/pypi_response/pytest_simple.html does contain data-core-metadata and not data-dist-info-metadata, so we should indeed account for the change in micropip/externals/mousebender/simple.py.
> 
> I just checked the test file and it seems like it contains both `data-core-metadata` and `data-dist-info-metadata`.

Oops, sorry! Yes, PyPI contains both the new key and the fallback. I was searching for `data-distinfo-metadata`, without the hyphen in `dist-info` – I stand corrected.


> > Unrelated point, but similar to @Carreau's suggestion above: we should remove the existing wheel files from tests/test_data/wheel/, since they will only bloat the repository as we proceed to add more commits to the default branch. They can be stored elsewhere
> 
> Yes, maybe let's discuss it in a separate issue. I don't like to put bunch of big files in the repository for testing, but I am also not sure whether putting it in a separate place is a good way to go.

Yes, opened a separate issue: #165

Please feel free to re-request a review once you are done here, thanks!

ryanking13:
@agriyakhetarpal Thanks, I think it is ready now.

micropip/package_index.py
```diff
@@ -153,6 +154,11 @@ def _compatible_wheels(
             hashes = file["digests"] if "digests" in file else file["hashes"]
             sha256 = hashes.get("sha256")
 
+            # Check if the metadata file is available (PEP 658 / PEP-714)
+            core_metadata: DistributionMetadata = file.get("core-metadata") or file.get(
+                "data-dist-info-metadata"
+            )
+
```
agriyakhetarpal, line 161: Question: should this be `"data-core-metadata"` (might have been missed)? The keys for the HTML API are prefixed with `"data-"`, but those for the JSON API are not.
```suggestion
            # Check if the metadata file is available (PEP 658 / PEP-714)
            core_metadata: DistributionMetadata = file.get("data-core-metadata") or file.get(
                "data-dist-info-metadata"
            )
	
```

agriyakhetarpal (approved):
Thanks, looks great, @ryanking13!

micropip/package_index.py
ryanking13: Good question. The `mousevender` renames `data-core-metadata` into `core-metadata` when loading HTML to Python dict so that both JSON and HTML responses can access the same key.

PR #141, opened 2024-10-10
Use singleton package manager for global API

ryanking13:
Create a singleton PackageManager class and expose its methods as a global APIs.

Thanks to @RulerOfCakes, all APIs except for uninstall are ported to the class method. I'll continue porting and cleaning up the remaining works.


micropip/__init__.py
```diff
@@ -1,19 +1,24 @@
...
-)
 from ._commands.uninstall import uninstall
+from .package_manager import PackageManager
 
 try:
     from ._version import __version__
 except ImportError:
     pass
 
+singleton_package_manager = PackageManager()
```
hoodmane, line 9: Maybe `_package_manager_singleton`?

ryanking13: Sounds good.

hoodmane (approved):
Thanks!

PR #142, opened 2024-10-18
More debug logging

Carreau:
Basically still trying to make the proxy works, but so many errors are swallowed because everything that triggers a ValueError is just ignore. So add a bunch of debug to at least be able to debug.

    async def add_requirement_inner(
        self,
        req: Requirement,
    ) -> None:
        ....
        try:
            if self.search_pyodide_lock_first:
                ...
            else:
                try:
                    await self._add_requirement_from_package_index(req)
                except ValueError:
                    # basically this swallow things it should not, like
                    improper content type and likely a bunch of other.

This is where many ValueError are swallowed, I will send more changes to
change various error types.

### DIFF
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import string
 import sys
 from collections import defaultdict
@@ -19,6 +20,8 @@ DEFAULT_INDEX_URLS = ["https://pypi.org/simple"]
 INDEX_URLS = DEFAULT_INDEX_URLS
 
 _formatter = string.Formatter()
+
+logger = logging.getLogger("micropip")
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


hoodmane:
Logging changes look good.

We seem to raise a LOT of `ValueError`s. The comment in transaction.py says: `If
the requirement is not found in package index` which means that it's expecting
to catch not found packages. Maybe make:
```py
class NotFoundError(Exception):
    pass

class NoSuchPackage(NotFoundError):
    pass

class NoCompatibleWheel(NotFoundError):
   pass
```
and then raise these in `querty_package` and `find_wheel` and switch the two `except ValueError` spots to catch `NotFoundError`?

Carreau:
I'm not 100% sure about the hierarchy, the problem is (to me) that `NotFoundError`, strongly screams 404 error, and AFAICT in your example `NoSuchPackage` and `NotFoundError` would then be the same.

And `NoCompatibleWheel` is really not a subclass of `NotFound`.


Maybe I'll split the logging into it's own things and then refactor the various errors.

Carreau:
I removed the change of exceptions and will send a subsequent PR.

micropip/package_index.py
```diff
@@ -224,7 +227,11 @@ def _select_parser(content_type: str, pkgname: str) -> Callable[[str], ProjectIn
             return ProjectInfo.from_simple_json_api
         case "application/json":
             return ProjectInfo.from_json_api
-        case "application/vnd.pypi.simple.v1+html" | "text/html":
+        case (
+            "application/vnd.pypi.simple.v1+html"
+            | "text/html"
+            | "text/html; charset=utf-8"
```
hoodmane, line 233: I guess there is also this change, but it seems minor enough and correct enough to merge without worrying too much.

hoodmane (approved):

hoodmane:
Thanks!

Carreau:
Yep, except on Python 3 you need  parentheses :-)

PR #143, opened 2024-10-18
Differenctiate `ValueError`s

Carreau:
ValueErrors are a bit too generic, and by raising/catching them we loose a potential useful information; For example an invalid content type for the index, is dealt as the same way than a missing wheel.

This introduces new exceptions for various cases, ant stop catching the now UnsupportedContentTypeError, as I believe it should be a hard error as we are likely having either an error with the index software, or a user configuraition error.

See also beginning of discussion in #142

-- 

Opening as draft, I'm expecting the tests to fails

### DIFF
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -218,6 +218,10 @@ def _contain_placeholder(url: str, placeholder: str = "package_name") -> bool:
     return placeholder in fields
 
 
+class UnsupportedContentTypeError(Exception):
+    pass
+
+
 def _select_parser(content_type: str, pkgname: str) -> Callable[[str], ProjectInfo]:
     """
     Select the function to parse the response based on the content type.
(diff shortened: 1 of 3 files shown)


micropip/package_index.py
```diff
@@ -234,7 +238,13 @@ def _select_parser(content_type: str, pkgname: str) -> Callable[[str], ProjectIn
         ):
             return partial(ProjectInfo.from_simple_html_api, pkgname=pkgname)
         case _:
-            raise ValueError(f"Unsupported content type: {content_type}")
+            raise UnsupportedContentTypeError(
+                f"Unsupported content type: {content_type}"
+            )
+
+
+class NoValidIndexForPackageError(Exception):
```
hoodmane, line 246: Maybe `PackageNotFoundError` would be better? `NoValidIndex` is a bit confusing IMO.

micropip/package_index.py
```diff
@@ -218,6 +218,10 @@ def _contain_placeholder(url: str, placeholder: str = "package_name") -> bool:
     return placeholder in fields
 
 
+class UnsupportedContentTypeError(Exception):
```
ryanking13, line 221: Maybe let's put the custom exceptions into a separate file (`error.py`) and import from there.

micropip/package_index.py
Carreau: That would be a synonym of PackageNotFoundError from importlib.metadata which is used un the same file... I've used a slightly longer name to differentiate.

PR #145, opened 2024-10-21
Add implied assertion in WheelInfo

Carreau:
Some simple indexes (anaconda), only list the full-pathname of download URL (assuming same domain, which make sens).

WheelInfo.download does not work on this case at it tries to make the requests on the current page domain.

Thus we implicitely assume that URL start with http (well https:// would be better, but when you prototype local index could be http.)

The error can be weird if we let it propagate (bad ziplife as you try to decode often a 404 html page as ZIP).

This thus just add an assert at construction time to catch the error early.

It should in the end be pushed earler in the code (likely at parsing time), where we are likely to know the index URL and be able to resolve URLs at that time.

### DIFF
--- a/micropip/wheelinfo.py
+++ b/micropip/wheelinfo.py
@@ -54,6 +54,9 @@ class WheelInfo:
     _dist_info: Path | None = None
 
     def __post_init__(self):
+        assert (
+            self.url.startwith(p) for p in ("http:", "https:", "emfs:", "file:")
+        ), self.url
         self._project_name = safe_name(self.name)
 
     @classmethod
@@ -63,6 +66,8 @@ class WheelInfo:
         See https://www.python.org/dev/peps/pep-0427/#file-name-convention
         """
         parsed_url = urlparse(url)
+        if parsed_url.scheme == "":
+            url = "file:///" + url
         file_name = Path(parsed_url.path).name
         name, version, build, tags = parse_wheel_filename(file_name)
         return WheelInfo(
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -257,7 +257,7 @@ async def test_fetch_wheel_fail(monkeypatch, wheel_base):
 
     msg = "Access-Control-Allow-Origin"
     with pytest.raises(ValueError, match=msg):
-        await micropip.install("htps://x.com/xxx-1.0.0-py3-none-any.whl")
+        await micropip.install("https://x.com/xxx-1.0.0-py3-none-any.whl")
 
 
 @pytest.mark.skip_refcount_check
--- a/tests/test_transaction.py
+++ b/tests/test_transaction.py
@@ -7,23 +7,29 @@ from packaging.tags import Tag
     "path",
     [
         SNOWBALL_WHEEL,
-        f"/{SNOWBALL_WHEEL}" f"a/{SNOWBALL_WHEEL}",
+        f"/{SNOWBALL_WHEEL}",
+        f"a/{SNOWBALL_WHEEL}",
         f"/a/{SNOWBALL_WHEEL}",
         f"//a/{SNOWBALL_WHEEL}",
     ],
 )
-@pytest.mark.parametrize("protocol", ["https:", "file:", "emfs:", ""])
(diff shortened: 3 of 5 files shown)


hoodmane:
Thanks!

hoodmane:
Looks like this gets hit in many tests though.

Carreau:
Hum, I'll dive into it more.

tests/test_install.py
```diff
@@ -257,7 +257,7 @@ def _mock_fetch_bytes(arg, *args, **kwargs):
 
     msg = "Access-Control-Allow-Origin"
     with pytest.raises(ValueError, match=msg):
-        await micropip.install("htps://x.com/xxx-1.0.0-py3-none-any.whl")
```
Carreau, line 260: It's not entirely clear if the typo here was on purpose.

Carreau:
Worse case I'll push a better error in `def download(self,...)`, but I prefer early failure when possible.

micropip/wheelinfo.py
```diff
@@ -63,7 +66,10 @@ def from_url(cls, url: str) -> "WheelInfo":
         See https://www.python.org/dev/peps/pep-0427/#file-name-convention
         """
         parsed_url = urlparse(url)
+        if parsed_url.scheme == "":
+            url = "file:///" + url
         file_name = Path(parsed_url.path).name
+        assert file_name, (url, file_name, parsed_url)
```
Carreau, line 72:
```suggestion
```

tests/test_transaction.py
```diff
@@ -7,17 +7,21 @@
     "path",
     [
         SNOWBALL_WHEEL,
-        f"/{SNOWBALL_WHEEL}" f"a/{SNOWBALL_WHEEL}",
```
Carreau, line 10: I think this is a mistake, and a `,` is needed.

tests/test_transaction.py
```diff
@@ -7,17 +7,21 @@
...
+        f"/{SNOWBALL_WHEEL}",
+        f"a/{SNOWBALL_WHEEL}",
         f"/a/{SNOWBALL_WHEEL}",
         f"//a/{SNOWBALL_WHEEL}",
     ],
 )
-@pytest.mark.parametrize("protocol", ["https:", "file:", "emfs:", ""])
+@pytest.mark.parametrize(
+    "protocol",
+    ["https:", "file:", "emfs:", ""],
```
Carreau, line 18:
```suggestion
    ["http:", "https:", "file:", "emfs:", ""],
```

tests/test_install.py
hoodmane: Looks pretty accidental to me. If it were on purpose you would expect a comment.

tests/test_transaction.py
```diff
@@ -7,17 +7,21 @@
...
+        f"/{SNOWBALL_WHEEL}",
+        f"a/{SNOWBALL_WHEEL}",
         f"/a/{SNOWBALL_WHEEL}",
         f"//a/{SNOWBALL_WHEEL}",
     ],
 )
-@pytest.mark.parametrize("protocol", ["https:", "file:", "emfs:", ""])
+@pytest.mark.parametrize(
+    "protocol",
+    ["http:", "https:", "file:", "emfs:", ""],
```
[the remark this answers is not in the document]
hoodmane: It stands for "Emscripten file system". It's needed because `file:` points to files in the actual file system. Probably could use some comments and better documentation.

tests/test_transaction.py
```diff
@@ -7,17 +7,21 @@
...
+)
 def test_parse_wheel_url1(protocol, path):
     pytest.importorskip("packaging")
     from micropip.transaction import WheelInfo
 
-    url = protocol + path
+    url = protocol + path if protocol else "file:///" + path
```
hoodmane, line 24: Why not put `file:///` in the parametrize decorator and do:
```suggestion
    url = protocol + path
```

hoodmane (approved):
Thanks for the maintenance work @Carreau! I have one suggestion otherwise lgtm.

PR #149, opened 2024-10-24
Do not print fill WheelInfo in debug.

Carreau:
And also do not add `_data` to the WheelInfo REPR.

Trying to print the full wheelinfo (with all the binary data), will hang the console it's way to much info.

So exclude the data from the repr, and only print the filename in debug login installing.


hoodmane (approved):
Thanks!

PR #151, opened 2024-10-31
Do not store pypi_response and metadata as gz.

Carreau:
This uncompress all files in `pypi_response` and update the relevant
section of test code to deal with the new format.

The idea is that binary blobs as test data are frowned upon for various
reasons, even more recenlty with the `xz` fiasco where there were used
to carry a backdoor.

Compression does not really reduce repository size as git is compressed
anyway;

And uncompressed files are harder to modify/review modifications.

-- 

This is done un 3 commits for logical consistency; in particular the second commits show the modifications of the gz files I did in #150 

I understand that this can/could be controversial as it's a lot of change, so I have no issues if this waits a bit and/or is closed later.


Carreau:
See [XZ backdoor incident on wikipedia for example](https://en.wikipedia.org/wiki/XZ_Utils#Backdoor_incident)

ryanking13 (approved):
Thanks!

PR #153, opened 2024-11-03
TST Test micropip.freeze() with pyodide-lock

ryanking13:
Add a new test case to test the output of micropip.freeze() with PyodideLockSpec defined in pyodide-lock.

Related: #107


hoodmane:
Thanks!

PR #155, opened 2024-11-10
Fix regression in loading package from local filesystem in Node

ryanking13:
There is a bug in micropip 0.7.0 that causes Node to fail to load packages from the local filesystem.

I also enabled Node CI to detect this kind of bug.


hoodmane (approved):
Thanks!

ryanking13:
Thanks for the review!

PR #159, opened 2024-11-22
Hardcode support for platform pyodide_wasm32

Carreau:
Also cahnge some Generators to Tuple, as it is not the first time that while debugging I exaust the generator when printing logs; and this make micropip installing the package to fail.

I don't like it; this feels like a hack, but it works, and I woudl be happy to get more guidance.

Also one weird thing, is if I installl pywavelets nightly, it installs numpy stable; while the index does have a numpy nightly;

And with the same config, if I ask it to install numpy; it does install numpy nightly.

So there is something wonkey in recursive dependency handling of indexes.

### DIFF
--- a/micropip/_utils.py
+++ b/micropip/_utils.py
@@ -2,7 +2,7 @@ import functools
 import json
 from importlib.metadata import Distribution
 from pathlib import Path
-from sysconfig import get_platform
+from sysconfig import get_config_var, get_platform
 
 from packaging.requirements import Requirement
 from packaging.tags import Tag
@@ -62,8 +62,15 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
 
 
 @functools.cache
-def sys_tags() -> list[Tag]:
-    return list(sys_tags_orig())
+def sys_tags() -> tuple[Tag, ...]:
+    new_tags = []
+    abi_version = get_config_var("PYODIDE_ABI_VERSION")
+    pyodide_platform_tag = f"pyodide_{abi_version}_wasm32"
+    for tag in sys_tags_orig():
+        if "emscripten" in tag.platform:
+            new_tags.append(Tag(tag.interpreter, tag.abi, pyodide_platform_tag))
+        new_tags.append(tag)
+    return tuple(new_tags)
 
 
 @functools.cache
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -214,6 +214,9 @@ def _fast_check_incompatibility(filename: str) -> bool:
     if not filename.endswith(".whl"):
         return False
 
+    if filename.endswith("wasm32.whl") and sys.platform == "emscripten":
+        return True
+
     if sys.platform not in filename and not filename.endswith("-none-any.whl"):
         return False
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+ - Fix a bug that prevented some wasm32 wheel to be recognized as compatible with pyodide
+  [#159](https://github.com/pyodide/micropip/pull/159)
+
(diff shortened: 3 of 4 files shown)


Carreau:
You can test this patch with :


```
import logging
log = logging.getLogger('micropip')
log.setLevel(10)
````

````
import micropip
await micropip.install('pywavelets', index_urls=['https://cors.carreau.workers.dev/scientific-python-nightly-wheels/simple'])
```

(installs numpy stable, and pywavelets nightly)


```
import micropip
await micropip.install('numpy', index_urls=['https://cors.carreau.workers.dev/scientific-python-nightly-wheels/simple'])
```

(installs numpy nightly)

Carreau:
(I probably need to add a test, but feedback first would be welcome).

ryanking13:
> Also one weird thing, is if I installl pywavelets nightly, it installs numpy stable; while the index does have a numpy nightly;
And with the same config, if I ask it to install numpy; it does install numpy nightly.
So there is something wonkey in recursive dependency handling of indexes.

Hmm, It must be a bug.

micropip/package_index.py
```diff
@@ -174,15 +174,15 @@ def _compatible_only(
         Checking compatibility takes a bit of time, so we use a generator to avoid doing it if not needed.
         """
 
-        releases_compatible = {
-            version: cls._compatible_wheels(files, version, name=name)
+        releases_compatible_x = {
+            version: tuple(cls._compatible_wheels(files, version, name=name))
```
ryanking13, line 178: I would like to keep the generator if possible, as checking compatibility of a long list of versions is redundant.

micropip/package_index.py
```diff
@@ -214,6 +214,9 @@ def _fast_check_incompatibility(filename: str) -> bool:
     if not filename.endswith(".whl"):
         return False
 
+    if filename.endswith("wasm32.whl") and sys.platform == "emscripten":
+        return True
```
ryanking13, line 218: Maybe `sys.platform` check is redundant, as we never expect micropip to be used in the non-Emscripten environment.

micropip/_utils.py
```diff
@@ -62,8 +62,13 @@ def get_files_in_distribution(dist: Distribution) -> set[Path]:
 
 
 @functools.cache
-def sys_tags() -> list[Tag]:
-    return list(sys_tags_orig())
+def sys_tags() -> tuple[Tag, ...]:
+    new_tags = []
+    for tag in sys_tags_orig():
+        if "emscripten" in tag.platform:
+            new_tags.append(Tag(tag.interpreter, tag.abi, "pyodide_2024_0_wasm32"))
```
ryanking13, line 69: I think you can use

```
import sysconfig
abi_version = sysconfig.get_config_var("PYODIDE_ABI_VERSION")
tag = f"pyodide_{abi_version}_wasm32"
```

instead, as abi version is expected to be changed when Pyodide ABI changes.

ryanking13 (commented):
Thanks, I left some comments. Could you also update the changelog?

micropip/package_index.py
hoodmane: At least not for installation. For making lock files, it would make sense.

ryanking13: > For making lock files

Do you mean using it under `pyodide venv` environment? I think it would still have `sys.platform == emscripten`.

ryanking13 (approved):
Thanks. Codewise looks good to me. But could you please add a testcase for the updated `sys_tags` function?

There is a test function [test_check_compatible](https://github.com/pyodide/micropip/blob/1fa0b523bf80014e6f27761efa5167709e46238f/tests/test_utils.py#L97C5-L97C26) but it runs under the host environment, hence no `PYODIDE_ABI_VERSION` config var exists.

So I would recommend to make a new test function that runs inside the Pyodide.

```py
@run_in_pyodide
def test_check_compatible_in_pyodide(selenium_standalone_micropip):
    import micropip
    # run test against check_compatible()
```

Carreau:
I've added a single test that the pywavelet wheel that was not installing is marked as compatible when run in pyodide; is that sufficient ?

micropip/package_index.py
Carreau: Well; at least some test locally runs under `sys.platform != emscripten`, and are useful for fast iteration.

PR #164, opened 2024-12-05
Add basic integration tests

ryanking13:
This adds basic integration tests for micropip, which queries to real world packages indexes (PyPI).

To prevent DoS, this is disabled by default in PR, and only run when `[integration]` is included in the PR or when the PR is merged and pushed to the main branch.

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -60,7 +60,7 @@ def test_integration_uninstall_basic(selenium_standalone_micropip, pytestconfig)
 
         micropip.uninstall("snowballstemmer")
 
-        packages = await micropip.list()
+        packages = micropip.list()
         assert "snowballstemmer" not in packages
 
     _run(selenium_standalone_micropip)
@@ -81,6 +81,6 @@ def test_integration_freeze_basic(selenium_standalone_micropip, pytestconfig):
         snowballstemmer.stemmer("english")
 
         lockfile = micropip.freeze()
-        assert "snowballstemmer" in json.loads(lockfile)
+        assert "snowballstemmer" in json.loads(lockfile)["packages"]
 
     _run(selenium_standalone_micropip)
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -5,12 +5,20 @@
 # To prevent sending many requests to remote servers, these tests are disabled by default.
 # To run these tests locally, invoke pytest with the `--integration` flag.
 
+import pytest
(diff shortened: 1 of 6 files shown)


tests/integration/test_integration.py
```diff
@@ -0,0 +1,79 @@
...
+# to test the micropip's ability to install packages from real world package indexes.
+
+# To prevent sending many requests to remote servers, these tests are disabled by default.
+# To run these tests locally, add `--integration` flag, when inovking pytest.
+
+from pytest_pyodide import run_in_pyodide
+
+
+def test_integration_install_basic(selenium_standalone_micropip, pytestconfig):
+    pytestconfig.getoption("--integration", skip=True)
```
hoodmane, line 12: Can we make a separate fixture or decorator for this so that we can say:
```py
@integration_only
@run_in_pyodide
def test_integration_install_basic(selenium_standalone_micropip):
    # test body
```
Or maybe it would be possible to skip the entire file instead of each test individually?

tests/integration/test_integration.py
```diff
@@ -0,0 +1,79 @@
+# integration tests for micropip
+# these test often requires querying to the real packages existing in PyPI,
+# to test the micropip's ability to install packages from real world package indexes.
+
+# To prevent sending many requests to remote servers, these tests are disabled by default.
+# To run these tests locally, add `--integration` flag, when inovking pytest.
```
hoodmane, line 6:
```suggestion
# To run these tests locally, add `--integration` flag, when invoking pytest.
```

tests/integration/test_remote_indexes.py
```diff
@@ -1,7 +1,7 @@
 # This file contains tests that actually query remote package indexes,
 # to ensure that micropip works with real-world package indexes.
 # Since running these tests will send many requests to remote servers,
-# these tests are disabled by default.
+# these tests are disabled by default in CI.
```
hoodmane, line 4: Above you said: "To run these tests locally, add `--integration` flag". Here you say they are disabled by default _in CI_. But aren't they are also disabled by default locally?

hoodmane (approved):
Thanks @ryanking13! I have some nits but generally looks good.

tests/integration/test_remote_indexes.py
ryanking13: Right, they are disabled both locally and CI by default. I'll update the wording

ryanking13:
Thanks for the review!

PR #168, opened 2024-12-15
Ruff rules for comprehensions and performance

cclauss:
% `ruff check --select=C4,C90,PERF --statistics`
```
3	C408   	[*] unnecessary-collection-call
3	C901   	[ ] complex-structure
3	PERF401	[ ] manual-list-comprehension
1	C403   	[*] unnecessary-list-comprehension-set
1	C419   	[*] unnecessary-comprehension-in-call
[*] fixable with `ruff check --fix`
```

In `micropip/externals/mousebender/simple.py`, Mypy is a bit behind pyupgrade on Python 3.12.

% `ruff rule C408`
# unnecessary-collection-call (C408)

Derived from the **flake8-comprehensions** linter.

Fix is always available.

## What it does
Checks for unnecessary `dict()`, `list()` or `tuple()` calls that can be
rewritten as empty literals.

## Why is this bad?
It's unnecessary to call, e.g., `dict()` as opposed to using an empty
literal (`{}`). The former is slower because the name `dict` must be
looked up in the global scope in case it has been rebound.

## Examples
```python
dict()
dict(a=1, b=2)
list()
tuple()
```

Use instead:
```python
{}
{"a": 1, "b": 2}
[]
()
```

## Fix safety
This rule's fix is marked as unsafe, as it may occasionally drop comments
when rewriting the call. In most cases, though, comments will be preserved.

## Options
- `lint.flake8-comprehensions.allow-dict-calls-with-keyword-arguments`

### DIFF
--- a/micropip/externals/mousebender/simple.py
+++ b/micropip/externals/mousebender/simple.py
@@ -5,9 +5,10 @@ import html
 import html.parser
 import urllib.parse
 import warnings
-from typing import Any, Literal, TypedDict
+from typing import Any, Dict, List, Optional, Union, Literal, TypeAlias, TypedDict
 
 import packaging.utils
+
 
 ACCEPT_JSON_V1 = "application/vnd.pypi.simple.v1+json"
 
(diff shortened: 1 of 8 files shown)


ryanking13 (approved):
Thanks. Looks good codewise, could you resolve the conflict?

cclauss:
Should we exclude `micropip/externals/mousebender/simple.py` given that it is external?

ryanking13:
@cclauss

Yes, please exclude the `external` directory. They are mostly copy/pasted.

ryanking13 (approved):
Thanks!

PR #177, opened 2025-01-23
Add support for constraints, install in topological order

### LINKED ISSUE #175: support constraints for transient dependencies?
## elevator pitch

Add the concept of [constraints] for controlling resolution of transient dependencies.

[constraints]: https://pip.pypa.io/en/stable/user_guide/#constraints-files

## motivation

For a number of use cases, constraints help achieve reproducible environments 
with more familiar tools and data than the still-prickly `pyodide-lock.json`.

## design ideas

For parity with the positional `dependencies`, this would likely be a list of PEP-508 specs which get normalized, either up-front affecting all future solves, or in a one-off `install`:

```py
micropip.set_constraints(["maybe-transient-dep==1.2.3"])

await micropip.install(
    ["direct-dep"], 
    constraints=[
        "another-transient-dep==4.5.6",
        "yet-another-transient-dep@https://example.com/yet_another_transient_dep.py3-none-any.whl",
    ]
)
```

In an outer build environment, the output of `pip list --format=freeze` could provide any "interesting" dependencies to something that should work in the browser, even if they need to be slightly tweaked to avoid any `emscripten`- or `pyodide`-specific `.whl`.

bollwyvl:
## references
- fixes #175

## changes
- [x] add `micropip.set_constraints`
- [x] add `constraints` to the signatures of `install` and `Transaction`
  - [x] per upstream, filter constraints that don't provide enough detail (e.g. Just A Wheel URL)
  - [x] merge specifiers to any requirement if found in the constraints (by canonical name)
- [x] rework `install` to handle constrained indirect dependencies of packages found in `pyodide-lock.json`
  - [x] add `requested_by` to the signatures of many `Transaction` methods
  - [x] capture the full as-requested `Transaction.dependency_graph`
  - [x] install packages in topological order based on graph

## breaking changes
- ~~hopefully none, as it's opt-in, and only adds keyword-only arguments to the already very finicky `install`, etc.~~
    - previously, _all_ packages from `pyodide-lock.json` were unconditionally installed before _any_ resolved wheels
    - now, all installations are handled the same way, waiting for any dependencies to be installed first
    - further, transactions containing cycles of dependencies will fail

### DIFF
--- a/docs/project/usage.md
+++ b/docs/project/usage.md
@@ -71,7 +71,8 @@ You can pass multiple packages to `micropip.install`:
 await micropip.install(["pkg1", "pkg2"])
 ```
 
-You can specify additional constraints:
+A dependency can specify be refined per the [PEP-508] spec:
+
 
 ```python
 await micropip.install("snowballstemmer==2.2.0")
@@ -79,10 +80,31 @@ await micropip.install("snowballstemmer>=2.2.0")
 await micropip.install("snowballstemmer[all]")
 ```
 
+[PEP-508]: https://peps.python.org/pep-0508
+
+### Disabling dependency resolution
+
 micropip does dependency resolution by default, but you can disable it,
 this is useful if you want to install a package that has a dependency
 which is not a pure Python package, but it is not mandatory for your use case:
 
 ```python
 await micropip.install("pkg", deps=False)
+```
+
+### Constraining indirect dependencies
+
+Dependency resolution can be further customized with optional `constraints`: these
+provide the versions (or URLs of wheels)
+
+```python
+await micropip.install("pkg", constraints=["other-pkg ==0.1.1"])
+```
+
+Default `constraints` may be provided to be used by all subsequent calls to
+`micropip.install`:
+
+```python
+micropip.set_constraints = ["other-pkg ==0.1.1"]
+await micropip.install("pkg")
 ```
--- a/docs/project/usage.md
+++ b/docs/project/usage.md
@@ -96,7 +96,7 @@ await micropip.install("pkg", deps=False)
 
 Dependency resolution can be further customized with optional `constraints`:
 these modify both _direct_ and _indirect_ dependency resolutions, while direct URLs
-in either a requirement or constraint will generally bypass any other specifiers.
+in either a requirement or constraint will bypass any other specifiers.
 
 As described in the [`pip` documentation][pip-constraints], each constraint:
 
@@ -108,6 +108,8 @@ As described in the [`pip` documentation][pip-constraints], each constraint:
     - a URL
   - _must not_  request any `[extras]`
 
+Multiple constraints of the same canonical name are merged.
+
 Invalid constraints will be silently discarded, or logged if `verbose` is provided.
 
 ```python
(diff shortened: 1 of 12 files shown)


hoodmane:
Looks nice so far I suppose the next step would be to actually use them.

docs/project/usage.md
```diff
@@ -71,18 +71,54 @@ You can pass multiple packages to `micropip.install`:
...
+        # package-with-no-version                  # missing version or URL
+    ]
+)
+```
+
+`micropip.set_constraints` replaces any default constraints for all subsequent
+calls to `micropip.install` that don't specify constraints:
+
+```python
+micropip.set_constraints = ["other-pkg ==0.1.1"]
```
ryanking13, line 121:
```suggestion
micropip.set_constraints(["other-pkg ==0.1.1"])
```

tests/test_install.py
```diff
@@ -99,6 +99,74 @@ def test_install_mixed_case2(selenium_standalone_micropip, jinja2):
     )
 
 
+PYTZ_2020_5_WHEEL = "pytz-2020.5-py2.py3-none-any.whl"
+PYTZ_2020_5_URL = f"https://files.pythonhosted.org/packages/89/06/2c2d3034b4d6bf22f2a4ae546d16925898658a33b4400cfb7e2c1e2871a3/{PYTZ_2020_5_WHEEL}"
```
ryanking13, line 103: Let's not use the external URL in the unittest.

You can put the wheel file in the [test directory](https://github.com/pyodide/micropip/tree/main/tests/test_data/wheel), and use the [wheel_catalog](https://github.com/pyodide/micropip/blob/5ce6e1f26f0680bf7db5f56bb9c98cbbe3726908/tests/conftest.py#L160C5-L160C18) fixture to generate a URL for the wheel. ([example](https://github.com/pyodide/micropip/blob/5ce6e1f26f0680bf7db5f56bb9c98cbbe3726908/tests/test_list.py#L45-L46))

micropip/transaction.py
```diff
@@ -36,14 +36,31 @@ class Transaction:
...
+                logger.debug("Transaction: discarding nameless constraint: %s", con)
+                continue
+            if con.extras:
+                logger.debug("Transaction: discarding [extras] constraint: %s", con)
+                continue
+            if not (con.url or len(con.specifier)):
+                logger.debug("Transaction: discarding versionless constraint: %s", con)
+                continue
+            con.name = canonicalize_name(con.name)
+            self.constrained_reqs[con.name] = con
```
ryanking13, line 62: Could you split this into a separate function and add some test?

ryanking13 (commented):
Thanks! I left some comment

tests/test_install.py
bollwyvl: hm, the wheel catalog is keyed by name, which makes it not great for testing sensitivity to version resolution. i'll see about the other stuff, and swing back to this,

micropip/transaction.py
bollwyvl: moved to `_utils`

tests/test_install.py
bollwyvl: using the pre-existing `pytest` fixture

docs/project/usage.md
bollwyvl: fixed

bollwyvl:
Had a thought which might skip all the bookkeeping: still do `gather` on dependencies of `pyodide-lock` packages, but then install PyPI wheels first... took a bit to get what's in there working, so will try a sibling PR...

bollwyvl:
I think #181 is the better way forward, but will keep this open as a draft for reference.

ryanking13:
Closing in favor of https://github.com/pyodide/micropip/pull/181

PR #181, opened 2025-01-27
Add support for constraints, install wheels before pyodide packages

### LINKED ISSUE #175: support constraints for transient dependencies?

bollwyvl:
## references
- fixes #175
- split off from #177

## changes

- [x] mostly as on #177
- [x] roll back the last few commits of #177
  - [x] `transaction.py`
    - [x] if constraints are present, keep gathering dependencies from Pyodide packages, but don't capture the entire graph
  - [x] `install.py`
    - [x] install all external wheels _before_ Pyodide packages
    - [x] avoids `loadPackage` magically fulfilling the dependencies which may be in `constraints`

## breaking changes
- this changes _many_ fewer signatures than #177
- there _might_ be cases in the wild where the previous behavior was somehow working


bollwyvl:
@ryanking13 @hoodmane this should Do The Thing without breaking as much!

CHANGELOG.md
```diff
@@ -6,6 +6,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- `micropip.install` now installs wheels from PyPI or custom indexes before built-in
+  Pyodide packages, reversing the previous behavior.
+  [#181](https://github.com/pyodide/micropip/pull/181)
```
ryanking13, line 13: Is there a user-facing change related to this? I think the order of installation is an implementation detail that is not visible to the user. What a user sees whether the whole set of packages (with dependencies) are installed or not.

bollwyvl: Yeah, I don't know: seems like In interactive package management, which can be modified by about five different flags now, I think "reversing what we used to do," warrants at least a note. I am _not sure_ if it's breaking, but things folk did before might not work now, and they might not need workarounds (like the double-install thing).

ryanking13: I see. Makes sense. Let me check the code more carefully and review.

I think the current wording could be a bit misleading to users though. When the same package exists in both the pyodide lock file and PyPI, the lockfile currently takes precedence (in terms of which package to choose, not the order of installation). But maybe the user will think it in a different way.

bollwyvl: Right: today, in the transaction phase once a dependency chain hits a pyodide package, further dependency resolution was skipped because the lockfile would take care of the rest. This was a performance win, for sure. In the install stage, they were all mixed together in `wheel_tasks`, and there was really no way to make any claims about what would happen.

With _this_ change, the transaction will keep going all the way down for every novel dependency.  In the install phase, the PyPI wheels get installed (in parallel) before calling `loadPackage`, which _won't_ re-install (and re-download) a pyodide wheel.

This _does_ mean there will be an extra `add_dependency_inner` for each `depends` item, but I couldn't figure out a way to do it that would be remotely correct.

bollwyvl: ... but that's a lot to say in a CHANGELOG, maybe there's a terser way?

bollwyvl:
Will work on merge conflicts.

CHANGELOG.md
bollwyvl: tried some more things on [92baa16](https://github.com/pyodide/micropip/pull/181/commits/92baa16724b95b947ab9da8866a287c192578218)

ryanking13: > With this change, the transaction will keep going all the way down for every novel dependency. In the install phase, the PyPI wheels get installed (in parallel) before calling loadPackage, which won't re-install (and re-download) a pyodide wheel.

Hmm, I still don't quite get what `which won't re-install` means. By default, during the dependency resolution step, micropip will check if the package is available in the lockfile, and select the one from the lockfile instead of the one from the PyPI.

So, in the installation step, we expect no overlap between `transaction.pyodide_packages` and  `transaction.wheels`. Does this PR change this behavior?

bollwyvl:
Merging `main`, looks like something is broken, but can't reproduce locally:

```    
    @pytest.mark.skip_refcount_check
    @run_in_pyodide(packages=["micropip"])
    async def test_load_binary_wheel2(selenium_standalone_micropip):
        from pyodide_js._api import repodata_packages
    
        import micropip
    
>       await micropip.install(repodata_packages.regex.file_name)

tests/test_install.py:329: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/lib/python3.12/site-packages/micropip/package_manager.py:146: in install
    ???
/lib/python3.12/site-packages/micropip/install.py:52: in install
    ???
/lib/python3.12/site-packages/micropip/transaction.py:71: in gather_requirements
    ???
/lib/python3.12/site-packages/micropip/transaction.py:84: in add_requirement
    ???
/lib/python3.12/site-packages/micropip/transaction.py:180: in add_requirement_inner
    ???
/lib/python3.12/site-packages/micropip/transaction.py:227: in _add_requirement_from_package_index
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   ValueError: Can't fetch metadata for 'regex-2024-9-11-cp312-cp312-pyodide-2024-0-wasm32-whl'. Please make sure you have entered a correct package name and correctly specified index_urls (if you changed them).
```

bollwyvl:
Ah, it looks like the `.whl` is being canonicalized as a package name into `-whl`
```
-wasm32-whl
```
... but also no idea why it's not failing locally...

bollwyvl:
Back to :heavy_check_mark:

CHANGELOG.md
ryanking13: Got it. I didn't consider putting multiple packages in `install(...)`. Thanks.

micropip/transaction.py
```diff
@@ -35,14 +43,23 @@ class Transaction:
...
             self.index_urls == package_index.DEFAULT_INDEX_URLS
         )
 
+        self.constrained_reqs, messages = validate_constraints(
+            self.constraints, self.ctx
+        )
+
+        if self.verbose and messages:
+            for constraint, msg in messages.items():
+                logger.info("Transaction: constraint %s discarded: %s", constraint, msg)
```
ryanking13, line 61: In the long run, it seems better to change it so that the installation fails when constraints resolution fails (just like pip's behavior). However, micropip's current dependency resolution is incomplete on its own, so let's leave it for now.

ryanking13 (approved):
Looks good to me overall. Thanks for your work!

bollwyvl:
Thanks all!

agriyakhetarpal:
Released in https://github.com/pyodide/micropip/releases/tag/v0.9.0, thank you!

PR #183, opened 2025-01-29
Test with Pyodide 0.27.2

agriyakhetarpal:
## Description

This PR bumps the Pyodide runtime version in our CI from 0.27.0a2 to 0.27.2. Fixes a couple of failing tests.

### DIFF
--- a/micropip/wheelinfo.py
+++ b/micropip/wheelinfo.py
@@ -69,8 +69,15 @@ class WheelInfo:
         See https://www.python.org/dev/peps/pep-0427/#file-name-convention
         """
         parsed_url = urlparse(url)
-        if parsed_url.scheme == "":
-            url = "file:///" + url
+        if not parsed_url.scheme:
+            parsed_url = ParseResult(
+                scheme="file",
+                netloc=parsed_url.netloc,
+                path=parsed_url.path,
+                params=parsed_url.params,
+                query=parsed_url.query,
+                fragment=parsed_url.fragment,
+            )
         file_name = Path(parsed_url.path).name
         name, version, build, tags = parse_wheel_filename(file_name)
         return WheelInfo(
--- a/micropip/wheelinfo.py
+++ b/micropip/wheelinfo.py
@@ -69,13 +69,15 @@ class WheelInfo:
         See https://www.python.org/dev/peps/pep-0427/#file-name-convention
         """
         parsed_url = urlparse(url)
-
-        # If parsed_url has no scheme, then add file:// to
-        # the URL and parse it again
         if not parsed_url.scheme:
-            url = "file:///" + url
-            parsed_url = urlparse(url, scheme="file")
-
+            parsed_url = ParseResult(
+                scheme="file",
+                netloc=parsed_url.netloc,
+                path=parsed_url.path,
+                params=parsed_url.params,
+                query=parsed_url.query,
+                fragment=parsed_url.fragment,
+            )
         file_name = Path(parsed_url.path).name
         name, version, build, tags = parse_wheel_filename(file_name)
         return WheelInfo(
--- a/micropip/wheelinfo.py
+++ b/micropip/wheelinfo.py
@@ -70,6 +70,7 @@ class WheelInfo:
         """
         parsed_url = urlparse(url)
         if not parsed_url.scheme:
+            url = "file:///" + url
             parsed_url = ParseResult(
                 scheme="file",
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
I managed to fix one out of two failing tests, and the integration tests haven't proceeded to run (maybe they should be their separate job instead of a step).

I don't fully understand the failure with `test_load_binary_wheel2` yet, but I can take a look at it in depth again later in the day. Please feel free to push to my branch here if you'd like. :)

agriyakhetarpal:
I managed to fix one out of two failing tests, and the integration tests haven't proceeded to run (maybe they should be their separate job instead of a step).

I don't fully understand the failure with `test_load_binary_wheel2` yet, but I can take a look at it in depth again later in the day.

ryanking13:
Thanks!

The failing test should use the fixture `selenium_standalone_micropip`, not `selenium`. Otherwise the micropip in the Pyodide distribution will be used. I think we should find all invalid fixtures and replace it.

> I don't fully understand the failure with test_load_binary_wheel2 yet, but I can take a look at it in depth again later in the day.

Looks like we are [appending `file:///` when no scheme is given](https://github.com/pyodide/micropip/blob/bebc19a2c8fb48ab58f79c13d4b801547a87bdf1/micropip/wheelinfo.py#L71-L74), but it does not update the scheme in the parsed_url, so micropip [fails here](https://github.com/pyodide/micropip/blob/bebc19a2c8fb48ab58f79c13d4b801547a87bdf1/micropip/wheelinfo.py#L189-L193) saying the scheme is None.

agriyakhetarpal:
Thanks for the pointer, @ryanking13! I fixed the tests, but I don't know if I did the change correctly.

micropip/wheelinfo.py
```diff
@@ -69,8 +69,15 @@ def from_url(cls, url: str) -> "WheelInfo":
...
-            url = "file:///" + url
+        if not parsed_url.scheme:
+            parsed_url = ParseResult(
+                scheme="file",
+                netloc=parsed_url.netloc,
+                path=parsed_url.path,
+                params=parsed_url.params,
+                query=parsed_url.query,
+                fragment=parsed_url.fragment,
+            )
```
ryanking13, line 80: I think we need to keep the `url = "file:///" + url` as well. 

How about:

```python
if parsed_url.scheme == "":
    url = "file:///" + url
    parsed_url = urlparse(url)
```

?

ryanking13 (approved):
Thanks! I left a comment about the scheme handling, otherwise looks good to me.

micropip/wheelinfo.py
agriyakhetarpal: 🤔 I think this is handled by `if not parsed_url.scheme`, right? Could you please clarify?

agriyakhetarpal:
`test_load_binary_wheel2` is failing consistently due to pickling errors, but I don't think it's related. I retried it a few times to see if it's not a flake, it isn't. Do you have any ideas, @hoodmane? I'll revert the change for the time being.

agriyakhetarpal:
I guess it was related, indeed, tests are passing now. I don't immediately see from https://github.com/pyodide/micropip/pull/183/commits/f1de916433914128a8308f7435b06e4eb6f03ede as to why.

micropip/wheelinfo.py
agriyakhetarpal: Ah, I did try this in this commit: https://github.com/pyodide/micropip/pull/183/commits/09772c810b3529751b5b1168169c5bcac59d1758, but it didn't seem to respect the URL scheme even when passing `scheme='file'` directly, so I reverted it.

agriyakhetarpal: Oh, I did try this in this commit: https://github.com/pyodide/micropip/pull/183/commits/09772c810b3529751b5b1168169c5bcac59d1758, but it didn't seem to respect the URL scheme even when passing `scheme='file'` directly and the tests failed, so I had to revert it. Did I miss anything?

ryanking13: Hmm, got it. That's weird. Thanks for checking! Probably we can improve how `url` and `parsed_url` are managed inside the `WheelInfo` class, but we can do that later.

agriyakhetarpal:
Thanks for the review! I've opened new issues so that we can track both items. I guess we should be fine to merge here, so I'll press the button.

PR #185, opened 2025-01-29
Move `mousebender` files to the `_vendored/` directory

agriyakhetarpal:
## Description

This PR moves the `mousebender` adaptation to the `vendored/` directory, as a follow-up PR to #178, as requested in https://github.com/pyodide/micropip/pull/178#discussion_r1931639510.


agriyakhetarpal:
Should `micropip/_vendored/mousebender/` contain an `__init__.py` file? My opinion is that it shouldn't, as we won't expect anyone to use `import micropip._vendored.mousebender` in their scripts, but we had `micropip/externals/__init__.py` so it would have been possible to use `import micropip.externals` prior to this PR.

I checked https://github.com/search?q=micropip.externals&type=code and the results are either not relevant or really old, from forks that were not updated (when `micropip` existed in the Pyodide repository and we included `micropip.externals._pip).

hoodmane:
Thanks!

agriyakhetarpal:
Thanks for the review!

ryanking13:
Thanks Agriya!

PR #186, opened 2025-01-29
Remove `pytest.importorskip("packaging")` usage for transaction/installation tests

agriyakhetarpal:
## Description

This is a remnant of #178. We can now remove the `pytest.importorskip("packaging")` calls, as the check is redundant and the packaging API is available.


hoodmane (approved):
Thanks!

agriyakhetarpal:
Thanks for the review!

PR #187, opened 2025-01-30
Fix regression when deps=False is set

### LINKED ISSUE #180: micropip.install() fails when deps=False
When specifying `deps=False`, I'm unable to install a number of packages. E.g.,:

```sh
await micropip.install("plotly", deps=False, verbose=True)

Collecting plotly
  Downloading plotly-5.24.1-py3-none-any.whl
Installing collected packages: plotly
  File "/lib/python3.12/site-packages/micropip/package_manager.py", line 133, in install
    return await install(
           ^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/install.py", line 92, in install
    await asyncio.gather(*wheel_promises)
  File "/lib/python3.12/site-packages/micropip/wheelinfo.py", line 126, in install
    raise RuntimeError(
RuntimeError: Micropip internal error: attempted to install wheel before downloading it?
```

ryanking13:
resolves #180

I forgot to call `await` when deps is not set.

- [x] changelog

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -39,7 +39,8 @@ def test_integration_install_no_deps(selenium_standalone_micropip, pytestconfig)
     async def _run(selenium):
         import micropip
 
-        await micropip.install("pytest-asyncio", deps=False)
+        # a version that is not in Pyodide distribution
+        await micropip.install("pytest-asyncio==0.24.0a1", deps=False)
 
         import pytest_asyncio  # noqa: F401
 
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -34,6 +34,28 @@ def test_integration_install_basic(selenium_standalone_micropip, pytestconfig):
 
 
 @integration_test_only
+def test_integration_install_no_deps(selenium_standalone_micropip, pytestconfig):
+    @run_in_pyodide
+    async def _run(selenium):
+        import micropip
+
+        await micropip.install("pytest-asyncio", deps=False)
+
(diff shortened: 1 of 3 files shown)


tests/integration/test_integration.py
```diff
@@ -33,6 +33,25 @@ async def _run(selenium):
...
+
+        await micropip.install("pyodide-micropip-test", deps=False)
+
+        try:
+            # pyodide-micropip-test depends on snowballstemmer
+            import snowballstemmer  # noqa: F401
+        except ModuleNotFoundError:
+            pass
+        else:
+            raise Exception("Should raise!")
```
agriyakhetarpal, line 50: As the intention here is that we want `import snowballstemmer` to not be installed due to `deps=False`, is there a better way to handle this?

i.e., is it possible to wrap it in a context manager with `with pytest.raises(ModuleNotFoundError` (or we have to do it in the current way because `pytest` isn't installed?)?

agriyakhetarpal (approved):
Thanks, @ryanking13! I have one question about the code, which is just for my understanding and won't block merging.

tests/integration/test_integration.py
ryanking13: > i.e., is it possible to wrap it in a context manager with with pytest.raises(ModuleNotFoundError (or we have to do it in the current way because pytest isn't installed?)?

Yeah, good point. `pytest.raises` would work too, but I wanted to keep the test simple, not loading any extra package.

ryanking13:
Thanks for the review!

PR #197, opened 2025-02-02
Unpin `pypa/build`, `pyodide-lock`, and `pytest` for testing requirements

agriyakhetarpal:
## Description

This pull request unpins:
- `pypa/build`, as we were using a very old version, and we were not the same version as what `pyodide-build` uses;
    - fixes compatibility with `build==1.2.2post1` API
- `pyodide-lock`, so that we can be aware when something upstream in `pyodide-lock` changes/fails here; and
- `pytest`, as v8 is reasonably stable now, and the latest version is 8.3.4 at the time of writing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ test = [
   "pytest-pyodide",
   "pytest-cov",
   "pytest<8.0.0",
-  "build==0.7.0",
+  "build",
   "pyodide-lock==v0.1.0a8",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ test = [
   "pytest-httpserver",
   "pytest-pyodide",
   "pytest-cov",
-  "pytest<8.0.0",
+  "pytest",
   "build",
   "pyodide-lock",
 ]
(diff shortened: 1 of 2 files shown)


ryanking13 (approved):
Thanks!

agriyakhetarpal:
Thanks for the review!

hoodmane:
I'm a bit uncomfortable about the idea that we might release pyodide-lock and only then discover that we broke micropip. But perhaps there isn't much api surface.

agriyakhetarpal:
I agree, @hoodmane – though my idea was that keeping `pyodide-lock` pinned would tell us about a failure in a new version only when we were to update it, and `pyodide-lock` isn't a dependency for `micropip` but an optional one. The same case happened for `build` here, which was fine because it didn't have a big API surface. If we keep it unpinned and a failure shows up in the next CI run, we can quickly patch it and fix the problem on the `pyodide-lock`/`micropip` side and exclude that `pyodide-lock` version in `micropip` via `pyodide-lock>=0.1.0a8; !=0.x.y`.

We could add a CI job to test with nightly versions of the test dependencies, but that could also be overkill.

hoodmane:
Yeah it just feels worse when we break ourselves. But I agree in practice it's the same. I'm sure the build 0.7.0 pin was accidentally inherited from the pyodide-build pin on build which used to access a very large number of build implementation details.

PR #206, opened 2025-02-11
ENH Add support for reinstalling packages (take 2)

### LINKED ISSUE #51: Ability to reinstall a different version of a package

ryanking13:
Resolve #51

Rebases #64.

This adds a new parameter, `reinstall` to the micropip.install API.

When the requested package is already installed in the system, and if it is not compatible:

- `reinstall` == False (default): raises error.
- `reinstall` == True: uninstall the installed package and install the new one.

image


micropip/install.py
```diff
@@ -60,16 +63,20 @@ async def install(
 
         pyodide_packages, wheels = transaction.pyodide_packages, transaction.wheels
 
-        package_names = [pkg.name for pkg in wheels + pyodide_packages]
+        packages_all = [pkg.name for pkg in wheels + pyodide_packages]
+        distributions = search_installed_packages(packages_all)
+
+        with indent_log():
+            uninstall_distributions(distributions, logger)
```
hoodmane, line 71: Shouldn't this only happen if `reinstall` is `True`?

ryanking13: The `distributions` parameter will always be an empty list if reinstall is false, as the transaction will not add any package that are already installed when reinstall is false.

But I think I can add that check just in case.

PR #208, opened 2025-02-15
FIX Respect yanked flag

ryanking13:
Resolve: #146 

Updates the resolver to respect `yanked` flag ([PEP-592](https://peps.python.org/pep-0592/)).

If the version with the yanked flag set is found, it will be considered as a last resort, when all other versions are incompatible.

- [x] changelog


micropip/transaction.py
```diff
@@ -327,6 +328,8 @@ def find_wheel(metadata: ProjectInfo, req: Requirement) -> WheelInfo:
         reverse=True,
     )
 
+    yanked_versions: list[list[WheelInfo]] = []
```
agriyakhetarpal, line 331: Should this be just `list[WheelInfo]`, similar to `wheels = list(releases[ver])` below? Creating a list from a list seems to be redundant here.

```suggestion
    yanked_versions: list[WheelInfo] = []
```

micropip/transaction.py
```diff
@@ -335,22 +338,42 @@ def find_wheel(metadata: ProjectInfo, req: Requirement) -> WheelInfo:
...
-        best_tag_index = float("infinity")
+        wheels = list(releases[ver])
+
+        # If the version is yanked, put it in the end of the candidate list.
+        # If we can't find a wheel that satisfies the requirement,
+        # install the yanked version as a last resort.
+        yanked = any(wheel.yanked for wheel in wheels)
+        if yanked:
+            yanked_versions.append(wheels)
+            continue
```
agriyakhetarpal, line 349: I think this control flow makes things a bit confusing for me, coupled with the fact that the variable defined above is named `wheels`. I understand that we are trying a two-pass approach by first looking for non-yanked wheels, and then searching within the yanked wheels if no non-yanked wheel in the first pass was found to be a suitable candidate for installation.

Instead of checking if a yanked wheel exists, we could create a smaller list using a list comprehension and manage its lifetime separately instead of branching out? i.e., using `yanked_wheels = [wheel for wheel in wheels if wheel.yanked]` and then passing that to `_find_best_wheel()` (which can return `None` if it receives `None` as an input).

tests/test_transaction.py
```diff
@@ -215,6 +215,42 @@ def test_find_wheel_invalid_version():
     assert str(wheel.version) == "0.15.5"
 
 
+def test_yanked_version():
+    from micropip._vendored.packaging.src.packaging.requirements import Requirement
+    from micropip.transaction import find_wheel
+
+    versions = ["0.0.1", "0.15.5", "0.9.1"]
```
agriyakhetarpal, line 222: I think we should add two more tests, one test for the case where one requests a yanked wheel explicitly (similar to the above point), and another case which checks if a warning was raised or not upon installing a yanked wheel (if you agree with my above suggestion, that is).

agriyakhetarpal (commented):
Thanks, @ryanking13! I have just a few comments, the code looks good to me. Again, thanks for the reminder on Discord.

micropip/package_index.py
```diff
@@ -163,6 +163,12 @@ def _compatible_wheels(
             # Size of the file in bytes, if available (PEP 700)
             # This key is not available in the Simple API HTML response, so this field may be None
             size = file.get("size")
+
+            # PEP-592:
+            # yanked can be an arbitrary string (reason) or bool.
+            # any string is considered as True, so we convert it to bool.
+            yanked = bool(file.get("yanked", False))
+
```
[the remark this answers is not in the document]
ryanking13: Sounds good to me. Thanks for the sugggestion

micropip/transaction.py
ryanking13: It is a list of version => list of wheels in that version.

tests/test_transaction.py
ryanking13: Thanks. Updated test cases.

micropip/transaction.py
agriyakhetarpal: Thanks for the explanation! This makes more sense to me. So, if I understood it correctly, we are checking for the version first and finding wheels from it (and then checking which is yanked) – rather than the other way around.

agriyakhetarpal (approved):
Thanks, @ryanking13! This looks great now.

I do wonder if Codecov is helping us, as you previously mentioned a few days ago somewhere (I can't find that comment) that it misses a few lines that are actually covered, as it can't run the tests in a WASM environment. For example, it reports that `YANKED_WARNING_MESSAGE` in line 244 of `micropip/transaction.py` was not covered, but it's definitely there in the `test_integration_install_yanked` case you added,

Nevertheless, maybe we can find a way to run `coverage.py` in a WASM environment and use the HTML/Markdown reports it can generate, as I don't think `sys.monitoring` needs any special attention in Pyodide. (Of course, not for this PR).

micropip/transaction.py
ryanking13: Yes, exactly.

ryanking13:
> Nevertheless, maybe we can find a way to run coverage.py in a WASM environment and use the HTML/Markdown reports it can generate, as I don't think sys.monitoring needs any special attention in Pyodide. (Of course, not for this PR).

Yes, I think running coverage inside WASM is not that hard. The hard part would be merging the results of the coverage.py inside and outside Pyodide.

ryanking13:
Thanks for the review!

PR #209, opened 2025-02-19
MAINT Rename repodata -> lockfile

ryanking13:
As title, after this change (and after releasing Pyodide), we can now drop the duplicated exports:

https://github.com/pyodide/pyodide/blob/41c53a05d3827883b64f48fd2ba3302c86413c93/src/js/types.ts#L469-L472


agriyakhetarpal (approved):
Thanks, @ryanking13!

PR #211, opened 2025-02-20
Fix packaging error after #163

ryanking13:
There was a bad merge


agriyakhetarpal (approved):
Oops, sorry for missing that and thanks for catching it, @ryanking13!

ryanking13:
It was my bad 😅. We were always load packaging when using `selenium_standalone_micropip` fixture (which was also fixed in this PR) so it was not detected.

PR #212, opened 2025-02-20
MAINT passthrough the `lockfile` object from the manager in freeze function

ryanking13:
Related: #112

Updates the `freeze` API to pass the lockfile into the child functions. No functional changes are included.

The goal of this PR is to gradually eliminate the `from ._compat import ...` syntax from non-manager codes. Instead, the manager object will drill in the compatibility layer to child functions.


micropip/freeze.py
```diff
@@ -64,3 +71,85 @@ def load_pip_package(dist: importlib.metadata.Distribution) -> dict[str, Any]:
...
+    lockfile_packages (dict):
+        The lockfile packages to check against.
+
+    extras (list):
+        List of extras for this package.
+
+    """
+    if lockfile_packages and package_name in lockfile_packages:
+        # don't check things that are in original repository
+        return
```
agriyakhetarpal, line 104:
```suggestion
    if package_name in lockfile_packages:
        # don't check things that are in original repository
        return
```

I assume that by checking for `package_name in lockfile_packages`, we will imply that `lockfile_packages` is truthy?

ryanking13: If lockfile_packages is None, `package_name in lockfile_packages` will raise error.

PR #224, opened 2025-03-26
ENH Add `extra_index_urls` and `index_strategy` parameters (DRAFT]

### LINKED ISSUE #223: My attempt at using anaconda.org PyPI index with micropip
### To reproduce

Following https://github.com/pyodide/pyodide/issues/4898#issuecomment-2637731077, I wanted to try out the CORS headers in anaconda.org so I have built a scikit-learn Pyodide wheel locally and uploaded it to anaconda.org. I was hoping to use the anaconda.org PyPI index https://pypi.anaconda.org/lesteve/simple like this:
```py
import micropip
await micropip.install("scikit-learn", index_urls="https://pypi.anaconda.org/lesteve/simple", pre=True)
```

### What I would have expected

scikit-learn dev wheel gets installed from my own index URL, but the dependencies (numpy, scipy, joblib, threadpoolctl) are installed from the lock-file

### What happens instead

Looking at the browser console, it looks like it is trying to find dependencies in https://pypi.anaconda.org/lesteve/simple but this index only has scikit-learn so this fails.

Python traceback:
```
  File "/lib/python3.12/site-packages/micropip/package_manager.py", line 133, in install
    return await install(
           ^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/install.py", line 53, in install
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 55, in gather_requirements
    await asyncio.gather(*requirement_promises)
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 62, in add_requirement
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 155, in add_requirement_inner
    await self._add_requirement_from_package_index(req)
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 214, in _add_requirement_from_package_index
    await self.add_wheel(wheel, req.extras, specifier=str(req.specifier))
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 271, in add_wheel
    await self.gather_requirements(wheel.requires(extras))
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 55, in gather_requirements
    await asyncio.gather(*requirement_promises)
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 59, in add_requirement
    return await self.add_requirement_inner(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 155, in add_requirement_inner
    await self._add_requirement_from_package_index(req)
  File "/lib/python3.12/site-packages/micropip/transaction.py", line 196, in _add_requirement_from_package_index
    metadata = await package_index.query_package(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/package_index.py", line 308, in query_package
    metadata, headers = await fetch_string_and_headers(url, _fetch_kwargs)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python3.12/site-packages/micropip/_compat/_compat_in_pyodide.py", line 63, in fetch_string_and_headers
    response = await pyfetch(url, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/lib/python312.zip/pyodide/http.py", line 449, in pyfetch
    raise AbortError(e) from None
pyodide.http.AbortError: NetworkError when attempting to fetch resource.
```

Part of the browser console that shows that micropip is trying to find dependencies in https://pypi.anaconda.org/lesteve/simple:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://pypi.anaconda.org/lesteve/simple/threadpoolctl/. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://pypi.anaconda.org/lesteve/simple/joblib/. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://pypi.anaconda.org/lesteve/simple/scipy/. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 404.
```

### More context

A simple work-around is to first install the dependencies without specifying a `index_urls` so that uses the lock-file I guess or the PyPI index for pure-Python wheels:
```py
import micropip
await micropip.install(["joblib", "threadpoolctl", "scipy"])
await micropip.install("scikit-learn", index_urls="https://pypi.anaconda.org/lesteve/simple", pre=True)
```

agriyakhetarpal:
Closes #223

### DIFF
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -268,10 +268,12 @@ def _select_parser(
             raise ValueError(f"Unsupported content type: {content_type}")
 
 
-async def query_package(
+# TODO: reduce mccabe complexity
+async def query_package(  # noqa: C901
     name: str,
     index_urls: list[str] | str,
     fetch_kwargs: dict[str, Any] | None = None,
+    strategy: str = "first-index",
 ) -> ProjectInfo:
     """
     Query for a package from package indexes.
@@ -286,6 +288,9 @@ async def query_package(
         it finds a package. If no package is found, an error will be raised.
     fetch_kwargs
         Keyword arguments to pass to the fetch function.
+    strategy
+        Index strategy to use when querying multiple indexes. The default is "first-index".
+        Valid values are: "first-index", "unsafe-first-match", and "unsafe-best-match".
     """
 
     _fetch_kwargs = fetch_kwargs.copy() if fetch_kwargs else {}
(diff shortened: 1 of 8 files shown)


agriyakhetarpal:
This PR has also been ready for review, and I can work on it further so that we can include it in an upcoming 0.10.0 release. I'll re-trigger the tests now. If I remember correctly, the test failures I had encountered were simple to fix – I would love feedback on the API design.

Please note that the behaviour deviates from `pip` and is closer to that of `uv`, because they have aligned themselves with security considerations better – see https://docs.astral.sh/uv/concepts/indexes/#searching-across-multiple-indexes for more information.

Therefore, I think it makes sense for us to be security-aware as well, given the inherent security model for running WASM in browsers and the fact that WASM as a format in general has been proven not to be particularly effective against, say, cryptocurrency mining malware.

micropip/package_manager.py
```diff
@@ -132,6 +136,29 @@ async def install(
             - If a list of URLs is provided, micropip will try each URL in order until \
             it finds a package. If no package is found, an error will be raised.
 
+        extra_index_urls:
```
ryanking13, line 139: I am not sure if we need to separate `extra_index_urls` as we already support multiple index_urls.

How about exposing an API to get the current index URLs for micropip instead?

For example,

```py
micropip.install("...", extra_index_urls=["A", "B", "C"])
```

can be replaced with

```py
current_index_urls = micropip.get_index_urls()
micropip.install("...", index_urls=current_index_urls + ["A", "B", "C"])
```

micropip/package_index.py
```diff
@@ -303,36 +308,101 @@ async def query_package(
...
-        try:
-            metadata, headers = await fetch_string_and_headers(url, _fetch_kwargs)
-        except HttpStatusError as e:
-            if e.status_code == 404:
-                logger.debug("NotFound (404) for %r, trying next index.", url)
+    projects_info = []
+
+    # With "first-index" strategy, we'll return the first match we find
+    # without checking the other indexes at all.
+    if strategy == "first-index":
```
ryanking13, line 315: Each strategy has duplicate logics, so I think we can separate this with multiple functions, extracting out common parts.

micropip/package_index.py
```diff
@@ -303,36 +308,101 @@ async def query_package(
...
+
+            content_type = headers.get("content-type", "").lower()
+            try:
+                base_url = urlunparse(urlparse(url)._replace(path=""))
+                parser = _select_parser(content_type, name, index_base_url=base_url)
+            except ValueError as e:
+                raise ValueError(f"Error trying to decode url: {url}") from e
+            return parser(metadata)
+    # With "unsafe-first-match" or "unsafe-best-match", we need to check all indexes
+    else:
```
ryanking13, line 343: I think it is still better to check the strategy string, as people might pass some strange value here. 
Or, we can probably check the value early and failfast.

micropip/package_index.py
```diff
@@ -303,36 +308,101 @@ async def query_package(
...
-        return parser(metadata)
-    else:
-        raise ValueError(
-            f"Can't fetch metadata for '{name}'. "
-            "Please make sure you have entered a correct package name "
-            "and correctly specified index_urls (if you changed them)."
-        )
+        # For "unsafe-first-match", return the first project that has matching versions
+        if strategy == "unsafe-first-match":
+            return projects_info[0]
```
ryanking13, line 383: Is it the right implementation? If the `projects_info[0]` does not have the compatible version, I think we should fallback to other indices.

The compatibility check is not handled in this function (at least in the current implementation). So I think so support this strategy, we'll need to modify the Transaction as well.

ryanking13 (commented):
Thanks for working on this Agriya! 

I am not sure about the `extra_index_urls` parameter, but having `index_strategy` sounds good to me. So how about separating the PR into two?

Also, I am not sure the current implementation is regarding the index strategy is correct, I left a comment about it,

PR #230, opened 2025-04-06
Fix relative index URLs

### LINKED ISSUE #229: Simple HTML repository relative URLs are broken
#150 (cc @Carreau) introduced support for relative index URLs. Unfortunately, the transformation from relative to absolute is broken:

https://github.com/pyodide/micropip/blob/d0d2e7ab4974b121964843143d5a2df9b6182a0a/micropip/package_index.py#L327

For an index at e.g. /main/pypa/simple/ and a package URL /main/pypa/simple/numpy/, the relative wheel file URL is made absolute to /, not to /main/pypa/simple/numpy/. This goes against HTML redirects and thus seems to go against the spec, which states (for the simple JSON spec which builds on the HTML spec) "If relative, they are relative to the current URL as if it were HTML."

Is there a specific index that relies on this broken redirect?

I will propose a PR to fix this.

juntyr:
Fixes #229

### DIFF
--- a/micropip/package_index.py
+++ b/micropip/package_index.py
@@ -324,9 +324,7 @@ async def query_package(
 
         content_type = headers.get("content-type", "").lower()
         try:
-            base_url = urlunparse(urlparse(url)._replace(path=""))
-
-            parser = _select_parser(content_type, name, index_base_url=base_url)
+            parser = _select_parser(content_type, name, index_base_url=url)
         except ValueError as e:
             raise ValueError(f"Error trying to decode url: {url}") from e
         return parser(metadata)


agriyakhetarpal (approved):
Thanks, @juntyr! Would it be possible for you to add a test?

ryanking13 (changes requested):
Hi, sorry for the late review.

Anaconda package index requires the previous behavior.

The index URL is:

- https://pypi.anaconda.org/pyodide/simple/

and the relative URL in the response is

- `/pyodide/simple/zstandard/0.22.0/zstandard-0.22.0-cp312-cp312-pyodide_2024_0_wasm32.whl`

So we needed to strip out the `path` part and append the relative URL.

Maybe there is some other package index which have different behavior? Could you please tell us which package index you are having an issue?

juntyr:
This is for the index I host together with my custom Pyodide deployment. The index lists all packages that were built just for Pyodide.

Each version of the lab has its own index, so e.g. there is

- https://lab.climet.eu/main/pypa/simple/ for the main branch version
- https://lab.climet.eu/v0.3.0/pypa/simple/ for v0.3.0

When navigating the links in the web browser, everything works, so the relative links seem to be correctly set up. From my understanding, the simple HTML index supports relative URLs with the same meaning as when using a web browser.

I could of course switch to absolute URLs but don't really want to bake in the full URL since it would directly tie everything to where the lab is currently being hosted.

ryanking13:
I see. Thanks.

So, if the relative URL starts with `/` (anaconda case), then it should be concatenated with the base URL, otherwise, it should be concatenated to the full URL.

It looks like `urllib.parse.urljoin` handles this correctly. Let me see if I can fix the logic with it.

```
>>> urljoin("https://my.index.com/simple/hello/", "../mywhl.whl")
'https://my.index.com/simple/mywhl.whl'
>>> urljoin("https://my.index.com/simple/hello/", "/mywhl.whl")
'https://my.index.com/mywhl.whl'
```

PR #235, opened 2025-05-26
Replace loadDynlibsFromPackage and get_dynlibs with `install` API

ryanking13:
This PR adopts `install` function (https://github.com/pyodide/pyodide/issues/4914), which is a stable API for installing packages that Pyodide runtime provides.

It replaces other compatibility functions such as `loadDynlibsFromPackage` and `get_dynlibs`, and removes duplicate codes.

### DIFF
--- a/micropip/_compat/_compat_not_in_pyodide.py
+++ b/micropip/_compat/_compat_not_in_pyodide.py
@@ -1,8 +1,10 @@
+import io
 import re
 from typing import IO, Any
 from urllib.error import HTTPError
 from urllib.request import Request, urlopen
 from urllib.response import addinfourl
+import zipfile
 
 from .compatibility_layer import CompatibilityLayer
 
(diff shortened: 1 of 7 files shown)


micropip/_compat/_compat_not_in_pyodide.py
```diff
@@ -52,14 +52,29 @@ async def fetch_string_and_headers(
...
+        Install a package from a buffer to the specified directory.
+        TODO: Remove host tests that depends on internal behavior of install (https://github.com/pyodide/micropip/issues/210)
+              to make the compat code simpler
+        """
+        from micropip.metadata import wheel_dist_info_dir
+
+        with zipfile.ZipFile(io.BytesIO(buffer)) as zf:
+            zf.extractall(install_dir)
+            normalized_name = filename.split("-")[0]
+            dist_dir = Path(install_dir) / wheel_dist_info_dir(zf, normalized_name)
```
agriyakhetarpal, line 71: Should the normalization happen as per https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization, rather than `split("-")[0]`?

agriyakhetarpal (commented):
Thanks, @ryanking13! I will leave it to @hoodmane for the final review here, as I might not be the best person to review this. I had a question, as indicated below:

hoodmane (approved):

hoodmane:
But the added test seems to be failing.

ryanking13:
Thanks for the review! The failing test was fixed in #238

PR #238, opened 2025-07-01
TST Fix `test_get_files_in_distribution`

ryanking13:
The recent pytest v8.4.1 includes some metadata file in the dist-info directory which is not included in the RECORD file. It was breaking our test.


agriyakhetarpal (approved):
This makes sense to me, thanks Gyeongjae!

PR #241, opened 2025-07-05
Replace relative URLs to absolute URL when freezing the lockfile

ryanking13:
This updates `micropip.freeze` to replace relative URLs (or just file names) to absolute URL.

This was discussed in https://github.com/pyodide/pyodide/pull/5652#issuecomment-2893609838 but I forgot to fix it before releasing 0.28.

This also fixes https://github.com/pyodide/pyodide/issues/5736

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        pyodide-version: ["0.28.0a2"]
+        pyodide-version: ["0.28.0"]
         test-config: [
             # FIXME: recent version of chrome gets timeout
             { runner: selenium, runtime: chrome, runtime-version: "125" },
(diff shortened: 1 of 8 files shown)


hoodmane (approved):
Thanks!

hoodmane:
Now we should merge this, release micropip, release the packages, update the package set, and backport the package set update.

ryanking13:
Yes.

ryanking13:
@alumni I cannot reproduce the behavior. Could you provide some reproducer?

ryanking13:
Oh, okay, thanks for the detailed reproducer.

So the problem is that `urljoin` behaves differently when the base path includes trailing slash or not.

```
>>> urljoin("file:///a/b/", "hello.py")
'file:///a/b/hello.py'
>>> urljoin("file:///a/b", "hello.py")
'file:///a/hello.py'
```

ryanking13:
https://github.com/pyodide/pyodide/pull/5763 will do the fix. I am not sure whether it will be fully compatible with the Windows path. But as far as I remember, node handles mixed slashes quite well.

PR #249, opened 2025-08-22
Add a subclass to optimize Versions

### LINKED ISSUE #73: Optimize the speed of dealing with Version
We parse and compare versions lots of times when searching for an appropriate package, but parsing and comparing Version (`packaging.version.Version`) is quite expensive.

See: https://github.com/pyodide/micropip/pull/71#issuecomment-1630859157

So it would be nice to do some micro-optimizations for handling versions, such as (https://github.com/pyodide/micropip/pull/71#discussion_r1259764043):

> If we want to go faster (no need to do it, just an idea). We can probably also subclass Version to something that has a frozen __str__ and __hash__ so it doesn't need to be re-computed at each dict comparison.

grden:
This PR adds class `CachedVersion`, a subclass of `packaging.version.Version` which caches its hash and string representations to avoid re-computation at each comparison.

This closes #73.

### DIFF
--- a/micropip/_cached_version.py
+++ b/micropip/_cached_version.py
@@ -2,22 +2,29 @@ from ._vendored.packaging.src.packaging.version import Version
 
 
 class CachedVersion(Version):
-    """This class is a subclass of Version with cached hash and string representations for performance."""
+    """This class is a subclass of Version with a lazily cached string representation for performance."""
 
-    __slots__ = ("_cached_hash", "_cached_str")
+    __slots__ = ("_cached_str",)
 
     def __init__(self, version: str) -> None:
         super().__init__(version)
-
-        # Cache expensive computations
-        self._cached_hash = hash(self._key)
-        self._cached_str = super().__str__()
-
-    def __hash__(self) -> int:
-        return self._cached_hash
+        self._cached_str = None
 
     def __str__(self) -> str:
+        if self._cached_str is None:
+            self._cached_str = super().__str__()
         return self._cached_str
+
+    @classmethod
+    def from_version(cls, version: Version) -> "CachedVersion":
+        """Helper method to create a CachedVersion from an existing Version object."""
+        if isinstance(version, cls):
+            return version
+        instance = cls.__new__(cls)
+        instance._version = version._version
+        instance._key = version._key
+        instance._cached_str = None
+        return instance
 
     def __repr__(self) -> str:
         """A representation of the CachedVersion that shows all internal state.
--- a/micropip/_cached_version.py
+++ b/micropip/_cached_version.py
@@ -1,27 +1,26 @@
-from ._vendored.packaging.src.packaging.version import Version, InvalidVersion
+from ._vendored.packaging.src.packaging.version import InvalidVersion, Version
 
 __all__ = ["CachedVersion", "parse_cached_version", "InvalidVersion"]
 
 
 class CachedVersion(Version):
-    """This class is a subclass of Version with cached hash and string representations for performance.
-    """
-    
-    __slots__ = ('_cached_hash', '_cached_str')
-    
+    """This class is a subclass of Version with cached hash and string representations for performance."""
+
+    __slots__ = ("_cached_hash", "_cached_str")
+
     def __init__(self, version: str) -> None:
         super().__init__(version)
-        
+
         # Cache expensive computations
         self._cached_hash = hash(self._key)
         self._cached_str = super().__str__()
-    
+
     def __hash__(self) -> int:
         return self._cached_hash
-    
+
     def __str__(self) -> str:
         return self._cached_str
-    
+
     def __repr__(self) -> str:
         """A representation of the CachedVersion that shows all internal state.
 
--- a/micropip/_cached_version.py
+++ b/micropip/_cached_version.py
@@ -1,0 +1,45 @@
+from ._vendored.packaging.src.packaging.version import Version, InvalidVersion
+
+__all__ = ["CachedVersion", "parse_cached_version", "InvalidVersion"]
+
+
+class CachedVersion(Version):
+    """This class is a subclass of Version with cached hash and string representations for performance.
+    """
+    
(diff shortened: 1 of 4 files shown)


micropip/package_index.py
```diff
@@ -39,7 +39,7 @@ class ProjectInfo:
     # List of releases available for the package, sorted in ascending order by version.
     # For each version, list of wheels compatible with the current platform are stored.
     # If no such wheel is available, the list is empty.
-    releases: dict[Version, Generator[WheelInfo, None, None]]
+    releases: dict[CachedVersion, Generator[WheelInfo, None, None]]
```
ryanking13, line 42: CachedVersion is a child class of the original Version, so I think the type hints do not need to change.

micropip/_cached_version.py
```diff
@@ -0,0 +1,44 @@
...
+        """A representation of the CachedVersion that shows all internal state.
+
+        >>> CachedVersion('1.0.0')
+        <CachedVersion('1.0.0')>
+        """
+
+        return f"<CachedVersion('{self}')>"
+
+
+def parse_cached_version(version_str: str) -> CachedVersion:
```
ryanking13, line 34: parse_cached_version simply calls`CachedVersion`, so I think we can remove it and simply call `CachedVersion(version)` directly.

micropip/package_index.py
```diff
@@ -9,11 +9,11 @@
 from typing import Any
 from urllib.parse import urljoin, urlparse
 
+from ._cached_version import CachedVersion, InvalidVersion, parse_cached_version
```
ryanking13, line 12: Was there some reason to import and re-export InvalidVersion inside the _cached version? I think we can keep importing it from the packaging module as-is.

micropip/_cached_version.py
```diff
@@ -0,0 +1,44 @@
+from ._vendored.packaging.src.packaging.version import InvalidVersion, Version
+
+__all__ = ["CachedVersion", "parse_cached_version", "InvalidVersion"]
```
ryanking13, line 3: We won't do `from _cached_version import *` so I think this `__all__` attribute in unnecessary.

micropip/_cached_version.py
ryanking13, line 1: This file needs some unittest

ryanking13 (commented):
Thanks, @grden for picking up this issue.

As this is related to the performance improvement, I think it would be good to have some simple benchmark to test if it really improves the performance.

It is not easy to test a real-like scenario as it depends on network conditions, etc. So I think writing a short script that simply compares the performance of Version and CachedVersion would be enough, without doing any network calls.

Could you write some benchmark script, run it in both native and Pyodide environment and see if you can observe performance gains?

micropip/_cached_version.py
grden: I've added tests in a separate file for now. Any feedback on the tests would be much appreciated, since I'm not too famililar with testing.

hoodmane:
Cc @henryiii @brettcannon does this optimization make sense to you? Would it make sense to contribute something like this upstream? I don't think there's anything special about our use case.

micropip/_cached_version.py
```diff
@@ -0,0 +1,29 @@
...
+class CachedVersion(Version):
+    """This class is a subclass of Version with cached hash and string representations for performance."""
+
+    __slots__ = ("_cached_hash", "_cached_str")
+
+    def __init__(self, version: str) -> None:
+        super().__init__(version)
+
+        # Cache expensive computations
+        self._cached_hash = hash(self._key)
```
hoodmane, line 13: Maybe we could compute these lazily rather than on construction?

ryanking13: +1

micropip/_utils.py
```diff
@@ -88,7 +89,11 @@ def parse_wheel_filename(
 
 # TODO: Move these helper functions back to WheelInfo
 def parse_version(filename: str) -> Version:
-    return parse_wheel_filename(filename)[1]
+    """Parses a version from a wheel filename, returning a CachedVersion."""
+    version_obj = parse_wheel_filename(filename)[1]
+    if isinstance(version_obj, CachedVersion):
+        return version_obj
+    return CachedVersion(str(version_obj))
```
ryanking13, line 96: parse_wheel_filename will always return `Version` object. Changing it to str and converting it to `CachedVersion` again is wasteful. How about creating a helper function, something like

```python
CachedVersion.from_version(version_obj)
```

ryanking13:
Thanks for writing the benchmark script @grden.

I left some comments, but overall it looks good. Let's wait for inputs from pypa folks as well.

---

The comment below is not directly related to your PR, but some ideas about optimizing this part more. Feel free to take a look if you are interested (no action need for this PR).

- Cashing the hash would be helpful when randomly accessing the dictionary that uses the version key, but one of the major operations against the Version object, I think, is `<=` or `>=` operations, which are not optimized by cashing the hash.

- Also, one bottleneck that I noticed when I was writing the original issue (#73) was that initializing the `Version` instance is quite heavy. So maybe we can think of making a pool that returns a shared `Version` instance... maybe something like

```python
global_version_pool: dict[str, Version] = {}

def parse_version(version: str) -> Version:
    global global_version_pool

    if version in global_version_pool:
        return global_version_pool[version]

    new_version = Version(version)
    global_version_pool[version] = new_version
    return new_version
```

micropip/_cached_version.py
brettcannon: I don't see the hash being a performance concern as it comes from a tuple which itself will cache the results.

https://github.com/pypa/packaging/blob/258202ed7f796bdb8a65252a66c3fbd3e69e97f6/src/packaging/version.py#L72-L73

ryanking13:
> This explains why I saw a significant speedup from string caching but less improvement from hash caching on previous benchmarks. I've updated the class to use lazy caching, and for __str__ only.

Hmm, but I don't think we call `str` against the Version object in package_index.py. So I am not sure caching str is that worth it for us.

Issue #205, opened 2025-02-08
Optional dependencies can not be combined with .whl URIs


chrysn:
To ease testing of unreleased versions, I'm publishing a wheel of any main branch build of my project aiocoap, so users can:

```python
>>> import micropip
>>> await micropip.install("https://raw.codeberg.page/aiocoap/aiocoap/@pages/dist/aiocoap-0.4.12.post0-py3-none-any.whl")
```

Frequently, users need to install dependencies, so my recommendation outside of development is that they do

```python
>>> import micropip
>>> await micropip.install("aiocoap[prettyprint,oscore]")
```

However, those can't be combined:

```python
>>> import micropip
>>> await micropip.install("https://raw.codeberg.page/aiocoap/aiocoap/@pages/dist/aiocoap-0.4.12.post0-py3-none-any.whl[prettyprint]")
Traceback (most recent call last):
[...]
packaging._tokenizer.ParserSyntaxError: Expected end or semicolon (after name and no valid version specifier)
    https://raw.codeberg.page/aiocoap/aiocoap/@pages/dist/aiocoap-0.4.12.post0-py3-none-any.whl[prettyprint]
         ^
```

(tested on https://pyodide.org/en/latest/console.html)

As far as I understand, this is a shortcoming of micropip, which is not present in pip (can't compare directly because AIU pip doesn't do wheels from URIs, but `pip install 'dist/aiocoap-0.4.12.post0-py3-none-any.whl[oscore]'` works).

Is this something that might be added, or is there a known workaround?

hoodmane:
It seems to me like a reasonable feature request, particularly since pip implements an analogous feature. It is a bit horrible to write `http://some-url.com/dist-name_plat.whl[extras]` but it's better than not having a way to do it at all. 

I can't say what workarounds people use maybe @agriyakhetarpal might know.

ryanking13:
Maybe we can allow putting extra suffix for `emfs://` or `file://` protocols only?

chrysn:
The interaction of those brackets with URIs is a bit odd in general -- and unlike pip, we're in an environment where explicit Python syntax is an option. So if combining this turns out to be problematic, it may be an option to introduce `await micropip.install("https://…/my-version.whl", ["feature1", "feature2"])`.

[...]

PR #253, opened 2025-09-15
install: Allow setting extras explicitly

chrysn:
This is a shot at https://github.com/pyodide/micropip/issues/205.

Frankly, I'm not super happy with it -- it's threading the extras quite wide through it, and it has the unsightly constraint that it only really works well with a single requirement (for it'd be quite unlikely that the same extras make sense for multiple requirements). I'll leave it here mainly as a discussion tool to find out whether https://github.com/pyodide/micropip/issues/205#issuecomment-3292353603 is really pointing in a good direction (and to see what CI has to say, for I have no clue how I'd test this a full pyodide around it).

### DIFF
--- a/micropip/package_manager.py
+++ b/micropip/package_manager.py
@@ -52,6 +52,7 @@ class PackageManager:
         constraints: list[str] | None = None,
         reinstall: bool = False,
         verbose: bool | int | None = None,
+        extras: set[str] | None = None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


ryanking13:
Thanks for the draft implementation. Do you think supporting `snowballstemmer[extras] @ https://.../snowballstemmer.*.whl` syntax would make things easier? If the implementation is cleaner, I think I am okay with that approach too.

PR #254, opened 2025-09-17
install: Support "pkg @ https://…/pkg.whl" syntax

chrysn:
The syntax "pkg @ https://…/pkg.whl" for a requirement to install (which comes from PEP-0508) is documented to be supported, but not actually implemented (instead, putting in just the URI is a micropip-proprietary supported feature, but it lacks support for features, see #205).

Open questions:
* Does it actually work? (I just tested parts on a local native Python build; let's see what CI says and maybe I can work off artifacts to actually test this)
* Should this kick off a deprecation process for the original syntax? (Probably not, as some documentation needs updating, but maybe a note in the code is due that if the one version causes trouble, go to the other).

Closes: https://github.com/pyodide/micropip/issues/205
Replaces: https://github.com/pyodide/micropip/pull/253


chrysn:
My testing procedure:

* `python3 -m build` and upload the wheel to my trusted CORS enabled location
* Go to https://pyodide.org/en/stable/console.html
* At browser console, do: `pyodide.loadPackage("https://downloads.endor.at/micropip-0.10.2.dev9+g19bdfe3b5-py3-none-any.whl")`
* In REPL, do `import micropip`; `await micropip.install("aiocoap[prettyprint] @ https://downloads.endor.at/aiocoap-0.4.14.post0-py3-none-any.whl")`
* `micropip.list()` shows pygments and other pretty-printing related packages

During testing, I had to rewrite things a bit, and ended up simplifying. I don't think that the previous code could ever do anything else but go for add_sheel, add_requirement_inner or raise (in particular because a specifier can only be present when there is no URL; essentially, the old code's re-parsing of req in the presence of a URL was what lost the information and caused #205 in the first place), but that'll be for reviewers (and tests) to check.

All in all I think that the code after this PR is even simpler than what was there before (let alone after what #253 would cause), and per @ryanking13's suggestion in https://github.com/pyodide/micropip/pull/253#issuecomment-3298757236, this would then be the route to go.

ryanking13:
Thanks. The `Requirement` class that packaging provides supports parsing the `name @ url` format. So I guess it would work, but we will need to carefully double check. I am glad that the change required was quite minimal.

micropip/transaction.py
```diff
@@ -80,21 +80,15 @@ async def add_requirement(self, req: str | Requirement) -> None:
...
-            if as_req.name and len(as_req.specifier):
-                return await self.add_requirement_inner(as_req)
-            if as_req.url:
-                req = as_req.url
-
-        if urlparse(req).path.endswith(".whl"):
-            # custom download location
-            wheel = WheelInfo.from_url(req)
-            check_compatible(wheel.filename)
-            return await self.add_wheel(wheel, extras=set(), specifier="")
```
ryanking13, line 95: I think we need to check the parsed url after converting the str to Requirement here. We don't support non `.whl` suffixes.

ryanking13: Ohh, okay. Thanks for the traceback. That makes sense.

ryanking13:
Hey @chrysn. I just wanted to let you know that I did not forget this issue. Looking at the error, I think there are a few other places that need to be modified. I will take a look when I have time.

PR #255, opened 2025-09-27
Remove importing IN_BROWSER from pyodide.ffi

ryanking13:
This is handled in [__init__.py](https://github.com/pyodide/micropip/blob/65f4c2f8af3c5d6873d15a2f7c1794013784e768/micropip/_compat/__init__.py#L7-L10) so additional check is unnecessary. I want to remove using this variable as it is not something I want to expose (https://github.com/pyodide/pyodide/pull/5916).


hoodmane:
Perhaps that `IN_BROWSER` should also be renamed and given an underscore? Also, does this not break pytest test collection?

hoodmane:
I guess if we run into trouble with test collection we can put it back.

hoodmane (approved):
Thanks!

ryanking13:
> Perhaps that IN_BROWSER should also be renamed and given an underscore?

Sounds good. Thanks.

> Also, does this not break pytest test collection?

I added pytest config to exclude compat directory from test collection, we don't have any doctest inside it so I think it is okay.

PR #257, opened 2025-10-03
Support extras in custom urls [integration]

### LINKED ISSUE #205: Optional dependencies can not be combined with .whl URIs

ryanking13:
Close #254 
Close #205 

This changes the behavior of micropip.install to install `extras` when the requirement is given in

```
pkg[extra] @ https://pkg.url.com/pkg.whl
```

format.

I noticed that we already support `pkg @ https://pkg.url.com/pkg.whl` syntax thanks to `packaging`, but we were not handling the extras correctly. This PR fixes it.

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -202,7 +202,7 @@ def test_install_url_based_wheel(selenium_standalone_micropip, pytestconfig):
         else:
             raise Exception("Should raise!")
 
-        await micropip.uninstall("typer")
+        micropip.uninstall("typer")
 
         await micropip.install(f"typer[all] @ {url}")
 


micropip/transaction.py
```diff
@@ -138,7 +144,7 @@ def check_version_satisfied(
                 "or micropip.uninstall(...) to uninstall the package first."
             )
 
-    async def add_requirement_inner(
+    async def add_requirement_inner(  # noqa: C901
```
hoodmane: Out of curiosity, what's C901?

Issue #237, opened 2025-06-27
Use dependency injection technique to manage compatibility layer


ryanking13:
Original comment: https://github.com/pyodide/micropip/pull/213#issuecomment-3012540003

Micropip has a compatibility layer, which defines a few functions or attributes that come from the Pyodide runtime.

In prod environment, the compat layer is imported from Pyodide, and for unit testing, the dummy compat layer is used ([code pointer](https://github.com/pyodide/micropip/blob/eefb57a70f7e187c4c9a1e9835ef2dc9f2dd370a/micropip/_compat/__init__.py#L9-L16))

For now, this compat layer is imported everywhere in micropip, but now we have a global singleton class for micropip (#141), I would like to make this compat layer a dependency of this class, using the dependency injection technique.

By doing this, we can

1. Remove `_compat_not_in_pyodide.py` from the production environment. ysing it only in unittest.
2. Users can modify the compatibility layer if needed. For example, if someone want to replace `fetch` with their custom method (downloading package from cache, indexDB, etc), it is possible by replacing the fetch method in compat layer.

---

I already made a few PRs to remove the importing compat layer globally, will continue working on it.

PR #263, opened 2025-11-23
MAINT Remove remaining codes that import compat functions directly

ryanking13:
Another step toward #237

This PR removes the all the remaining code that imports function directly from compatibility layer package. instead all the compatibiilty codes are injected as a parameter or a class attribute.

After this, we can clean up the `_compat_not_in_pyodide.py` from the micropip package wheel as they are not used in runtime at all.

The code structure is not ideal as we inject compat layer everywhere... but I think we can fix that later.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,11 +30,11 @@ jobs:
           ]
 
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
Oops, I totally missed your request for review here, @ryanking13. Please feel free to ping me on Discord at any time, always, if this happens again!

ryanking13:
> Oops, I totally missed your request for review here, @ryanking13. Please feel free to ping me on Discord at any time, always, if this happens again!

No worries. I also totally forgot about this PR 😅 As I am preparing for next Pyodide release, I was cleaning up some stale PRs.

PR #274, opened 2026-04-29
Fix docs mentioning that index URLs must support JSON API (they do not)

### LINKED ISSUE #273: Outdated docs for `index_urls`
Looking at the micropip [documentation](https://micropip.pyodide.org/en/latest/project/api.html), I believe the part regarding `index_urls` is outdated:

Image

My understanding is that since https://github.com/pyodide/micropip/issues/62 was merged, supporting the legacy JSON API is no longer a requirement to use `index_urls`.
After reading the [relevant code](https://github.com/pyodide/micropip/blob/main/micropip/package_index.py), my understanding is that the legacy JSON is not used at all by default.

I was experimenting with micropip and private package index, and one thing that might be interested to add in this portion of the docs: the index must have CORS headers (PyPI [does](https://github.com/pypi/warehouse/pull/13222/changes) since 2023), otherwise you get hard-to-debug errors.

alexprengere:
Fixes #273

### DIFF
--- a/micropip/package_manager.py
+++ b/micropip/package_manager.py
@@ -120,17 +120,21 @@ class PackageManager:
         index_urls:
 
             A list of URLs or a single URL to use as the package index when looking
-            up packages. If None, *https://pypi.org/pypi/{package_name}/json* is used.
+            up packages. If None, *https://pypi.org/simple* is used.
 
-            - The index URL should support the \
+            - The index URL may support the \
             `JSON API <https://warehouse.pypa.io/api-reference/json/>`__ .
 
-            - The index URL may contain the placeholder {package_name} which will be \
+            - The index URL may contain the placeholder {package_name}, for example \
+            *https://pypi.org/pypi/{package_name}/json*, this placeholder will be \
             replaced with the package name when looking up a package. If it does not \
             contain the placeholder, the package name will be appended to the URL.
 
             - If a list of URLs is provided, micropip will try each URL in order until \
             it finds a package. If no package is found, an error will be raised.
+
+            - The index URL must support CORS when used in a web browser. Note that \
+            *https://pypi.org/simple* has CORS enabled.
 
         constraints:
 
--- a/micropip/package_manager.py
+++ b/micropip/package_manager.py
@@ -122,19 +122,13 @@ class PackageManager:
             A list of URLs or a single URL to use as the package index when looking
             up packages. If None, *https://pypi.org/simple* is used.
 
-            - The index URL may support the \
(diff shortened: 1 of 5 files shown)


micropip/package_manager.py
```diff
@@ -120,18 +120,22 @@ async def install(
...
+            - The index URL may contain the placeholder {package_name}, for example \
+            *https://pypi.org/pypi/{package_name}/json*, this placeholder will be \
             replaced with the package name when looking up a package. If it does not \
             contain the placeholder, the package name will be appended to the URL.
 
             - If a list of URLs is provided, micropip will try each URL in order until \
             it finds a package. If no package is found, an error will be raised.
 
+            - The index URL must support CORS when used in a web browser. Note that \
+            *https://pypi.org/simple* has CORS enabled.
```
ryanking13:
```suggestion
            - The index URL must support CORS when used in a web browser.
```

Sounds redundant to me.

micropip/package_manager.py
```diff
@@ -120,18 +120,22 @@ async def install(
         index_urls:
 
             A list of URLs or a single URL to use as the package index when looking
-            up packages. If None, *https://pypi.org/pypi/{package_name}/json* is used.
+            up packages. If None, *https://pypi.org/simple* is used.
 
-            - The index URL should support the \
+            - The index URL may support the \
             `JSON API <https://warehouse.pypa.io/api-reference/json/>`__ .
```
ryanking13: Let's just say that the remote index should support [Simple repository API](https://packaging.python.org/en/latest/specifications/simple-repository-api/). We support JSON api but it is not standardized so I don't really want to let users know about it. We may drop supporting it later.

micropip/package_manager.py
```diff
@@ -120,18 +120,22 @@ async def install(
...
 
-            - The index URL may contain the placeholder {package_name} which will be \
+            - The index URL may contain the placeholder {package_name}, for example \
+            *https://pypi.org/pypi/{package_name}/json*, this placeholder will be \
             replaced with the package name when looking up a package. If it does not \
             contain the placeholder, the package name will be appended to the URL.
```
ryanking13: Let's also remove this as well.

alexprengere:
I pushed the suggestions. I also changed similar things in `micropip/package_manager.py`

hoodmane (commented):
Revert test formatting changes.

ryanking13:
> Revert test formatting changes.

I think they are done by pre-commit.ci for some reason

alexprengere:
Yeah I did not push those, and I did not configure precommit locally so this is coming from the CI

hoodmane (approved):
Thanks!

